PR #129, opened 2026-03-19
fix(fp-check): use correct JSON response format in stop hooks

jonathanreedmevs:
## Summary

- `Stop` and `SubagentStop` hook prompts were instructing Claude to return plain text (`'block'` or `'approve'`)
- Claude Code expects prompt-type hooks to respond with JSON, causing `Stop hook error: JSON validation failed` on every session end
- Updated both prompts to respond with `{"decision": "block", "reason": "..."}` to block, or `{}` to allow stopping

## Root Cause

The hook prompts on [line 10](https://github.com/trailofbits/skills/blob/19c463e3f628ed86c8e6befd142ce7d0653d8990/plugins/fp-check/hooks/hooks.json#L10) and [line 22](https://github.com/trailofbits/skills/blob/19c463e3f628ed86c8e6befd142ce7d0653d8990/plugins/fp-check/hooks/hooks.json#L22) ended with:

```
return 'block' with the specific gaps.
return 'approve'.
```

Claude Code parses the hook response as JSON. Plain text fails JSON validation, producing the error even though the hook logic itself was correct.

## Fix

Replace plain-text return instructions with explicit JSON response format matching the [Claude Code stop hook schema](https://code.claude.com/docs/en/hooks) (`decision` field). Bump version to `1.0.1` so installed clients receive the update.

```diff
- If ANY bug is missing ANY of these phases or the gate review, return 'block' with the specific gaps.
- If all bugs have complete verification with verdicts, return 'approve'.
- If the conversation is not about fp-check verification at all, return 'approve'.
+ If ANY bug is missing ANY of these phases or the gate review, respond with JSON: {"decision": "block", "reason": "<specific gaps>"}
+ If all bugs have complete verification with verdicts, respond with JSON: {}
+ If the conversation is not about fp-check verification at all, respond with JSON: {}
```

Same change applies to the `SubagentStop` prompt on line 22.

## Discovery

Bug discovered by Claude while debugging the `Stop hook error: JSON validation failed` error during active use of the fp-check skill.

## Bug Replication

- [ ] Start a Claude Code session with fp-check installed
- [ ] Prompt Claude with anything
- [ ] Confirm `Stop hook error: JSON validation failed` appears
- [ ] Apply fix, restart session, confirm error is gone


dguido (approved):
Bug confirmed real; response schema corrected to the documented ok/reason format and version bumped to 1.0.2. CI green.

PR #137, opened 2026-03-27
fix: fp-check Stop hooks return JSON instead of plain text

bluzername:
## Problem

The fp-check plugin's Stop and SubagentStop hooks cause "JSON validation failed" error every time Claude Code exits a session. The hooks use `type: "prompt"` which expects the LLM to return structured JSON, but the prompt instructions tell it to return plain text like `'approve'` or `'block'`.

I was getting this error on every session exit after installing fp-check and it was confusing because the plugin wasnt even being used in those sessions.

## Root Cause

In `plugins/fp-check/hooks/hooks.json`, both hooks have prompts ending with:

```
return 'block' with the specific gaps.
return 'approve'.
```

But Claude Code expects prompt-type hooks to return valid JSON like:
```json
{"decision": "approve"}
```

## Fix

Updated the return instructions in both hooks to ask for JSON format:

**Stop hook** (line 10):
```
return JSON: {"decision": "block", "reason": "<specific gaps>"}
return JSON: {"decision": "approve"}
```

**SubagentStop hook** (line 22):
Same pattern.

Also added `IMPORTANT: Return valid JSON only, no text outside the JSON object.` to both prompts to make sure the LLM doesnt add extra text around the JSON.

## Impact

This error happens on EVERY session exit when fp-check is installed, even for sessions that have nothing to do with false positive verification. The fix makes both hooks return proper JSON so Claude Code can parse the response correctly.

Closes #131


vishalverma8200 (approved):

nicobits:
Thank you for putting this together, it accurately targeted #131 and the writeup was really clear and helpful. The same bug ended up getting tackled via an earlier #129. Since both PRs fix hooks.json, will be closing as superseded. Thank you for your help getting this issue resolved!

PR #153, opened 2026-04-24
Refresh trailmark skills for public Trailmark 0.2.x

tob-scott-a:
Aligns all skills with the now-public trailmark package (pypi.org/project/trailmark, github.com/trailofbits/trailmark):

- Replace hardcoded language tables with runtime detection via trailmark.parse.detect_languages() and --language auto, so the skill never goes stale as Trailmark adds languages (21 supported as of 0.2.x: Python, JS/TS, PHP, Ruby, C/C++, C#, Java, Go, Rust, Solidity, Cairo, Circom, Haskell, Erlang, Miden Assembly, Swift, Objective-C, Kotlin, Dart)
- Remove phantom --passes CLI flag from trailmark-structural; pre-analysis is a QueryEngine.preanalysis() method, not a flag
- Use native trailmark diff --json in graph-evolution alongside the subgraph-diff helper script
- Add `finding` and `audit_note` annotation kinds to the documented list (set by augment_sarif / augment_weaudit)
- Document trailmark entrypoints and trailmark augment subcommands that exist in the public package
- Bump plugin to 0.8.1

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -225,7 +225,7 @@
     },
     {
       "name": "trailmark",
-      "version": "0.8.0",
+      "version": "0.8.1",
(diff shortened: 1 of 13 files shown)


tob-scott-a:
@claude review once

tob-scott-a:
> [!NOTE]
> Codex wrote this comment.

Addressed the inline review nits in 4a4e3aa:\n\n- switched the  and  snippets to pass the target path via argv instead of embedding  in Python source\n- removed stderr suppression from the native  example and added explicit non-empty diff-file guidance\n- synced  with the documented  /  annotation kinds\n- re-ran Validated 72 plugin skills against 73 Codex entries successfully. successfully

tob-scott-a:
> [!NOTE]
> Codex wrote this comment.

Addressed the inline review nits in `4a4e3aa`:

- switched the `detect_languages` and `QueryEngine` snippets to pass the target path via argv instead of embedding `{args}` in Python source
- removed stderr suppression from the native `trailmark diff` example and added explicit non-empty diff-file guidance
- synced `query-patterns.md` with the documented `FINDING` / `AUDIT_NOTE` annotation kinds
- re-ran `python3 .github/scripts/validate_codex_skills.py` successfully

PR #165, opened 2026-05-14
fix(fp-check): remove top-level Stop hook to stop firing on unrelated sessions (#143)

### LINKED ISSUE #143: fp-check Stop hook with matcher "*" fires on every unrelated session, causing forced LLM turns
## Summary

## Reproduction

## Why this is wrong

## Suggested fixes (any of)

\`\`\`bash
\`\`\`

## Affected file

## Environment

remiforall:
## Summary

Closes #143.

The `Stop` hook in `plugins/fp-check/hooks/hooks.json` is registered with `"matcher": "*"`, so it fires on **every** Claude Code session stop — including sessions that have nothing to do with fp-check verification. This burns a full 30-second LLM turn on every unrelated stop and pollutes the parent context with stop-feedback blocks.

The hook's own fallback rule (`"If the conversation is not about fp-check verification at all, return 'approve'"`) does not prevent the issue: the hook still fires, the model still runs, the turn is still consumed — just to discover that it should no-op.

## Fix

Removes the top-level `Stop` hook entirely (option 3 of three workarounds discussed in #143).

The companion `SubagentStop` hook stays. It is the right place to enforce structured-output completeness, because it only matches actual fp-check subagents (`data-flow-analyzer`, `exploitability-verifier`, `poc-builder`) and never fires on unrelated sessions. Verification rigor is preserved at the level where it is actionable.

## Impact

- Verification rigor: unchanged. `SubagentStop` continues to enforce phase-output completeness on every fp-check subagent.
- Unrelated sessions: no longer pay the cost of a full LLM turn at every `Stop`.
- Plugin version bumped from `1.0.0` to `1.0.1` (per repo CLAUDE.md guidance: substantive change to hook behavior).

## Validation

```
$ python3 .github/scripts/validate_codex_skills.py
Validated 73 plugin skills against 74 Codex entries successfully.
```

CODEOWNERS unchanged (`/plugins/fp-check/ @ahpaleus @dguido`).

## Test plan

- [x] Diff reviewed for completeness (3 files: hooks.json + 2 version bumps)
- [x] Codex skills validator passes
- [x] No CODEOWNERS change required
- [ ] CI green
- [ ] Reviewer ack from @ahpaleus or @dguido that removing the top-level Stop hook is acceptable (per @GeeksikhSecurity's own recommendation in #143: "Option 3 is the simplest and loses no real coverage.")

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -375,7 +375,7 @@
     },
     {
       "name": "fp-check",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "description": "Systematic false positive verification for security bug analysis with mandatory gate reviews",
       "author": {
         "name": "Maciej Domanski"
--- a/plugins/fp-check/.claude-plugin/plugin.json
+++ b/plugins/fp-check/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "fp-check",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Systematic false positive verification for security bug analysis with mandatory gate reviews",
   "author": {
     "name": "Maciej Domanski"
--- a/plugins/fp-check/hooks/hooks.json
+++ b/plugins/fp-check/hooks/hooks.json
@@ -1,18 +1,6 @@
 {
-  "description": "Enforce fp-check verification completeness — prevent premature verdicts and incomplete agent output",
+  "description": "Enforce fp-check subagent output completeness — verify data-flow-analyzer, exploitability-verifier, and poc-builder produced their required phase outputs",
   "hooks": {
-    "Stop": [
-      {
-        "matcher": "*",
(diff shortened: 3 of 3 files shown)


sandr8:
@remiforall can you please sign the CLA so that this PR can move along?

dguido:
Thanks for digging into #143 and confirming the repro, @remiforall — the diagnosis here is spot-on.

Closing in favor of #188, which fixes the same issue at the root cause. Rather than deleting the top-level `Stop` hook (which would leave the identical always-on bug in `SubagentStop`, and drop the overall verdict/gate-review completeness check on inline runs that don't dispatch subagents), #188 moves the hooks into skill/agent frontmatter so they're scoped to fp-check's lifecycle and only fire when the skill is actually in use. That fixes both hooks and keeps full coverage.

Your issue analysis directly shaped the fix — much appreciated.

PR #167, opened 2026-05-16
plugins/claude-in-chrome-troubleshooting: shorten skill dir to fix Codex 64-char name limit

siteexperts:
## Problem

Codex constructs each skill's internal name as `<plugin-dir>-<frontmatter.name>` and rejects anything over **64 chars**. This plugin's defaults produced a 65-char joined name:

```
plugins/claude-in-chrome-troubleshooting   (plugin dir, 32 chars)
+
SKILL.md frontmatter `name: claude-in-chrome-troubleshooting` (32 chars)
= claude-in-chrome-troubleshooting-claude-in-chrome-troubleshooting (65 chars)
```

Codex error in the user's log (verified on macOS, `~/.codex/log/codex-tui.log`):

```
ERROR codex_core::session: failed to load skill
  .../plugins/claude-in-chrome-troubleshooting/skills/troubleshooting/SKILL.md:
  invalid name: exceeds maximum length of 64 characters
```

Other single-skill trailofbits plugins with the plugin-name==skill-name pattern (e.g. `ask-questions-if-underspecified` at 63 chars) just barely fit; this one is two chars over.

## Fix

Two changes:

1. **commit dbcf09d** — Rename the inner `skills/<name>/` directory from `claude-in-chrome-troubleshooting` → `troubleshooting`. Tidies up the layout and matches the convention used by multi-skill plugins like `building-secure-contracts/skills/<distinct-skill-name>/`. Also re-points `.codex/skills/claude-in-chrome-troubleshooting` symlink at the new path.

2. **commit cbfd05f** — Rename the SKILL.md frontmatter `name:` field from `claude-in-chrome-troubleshooting` → `troubleshooting`. **This is the load-bearing fix**: Codex's joined-name calculation uses the frontmatter, not the dir name. Without this, the dir rename alone leaves the joined name at 65 chars.

After both changes, the joined name is `claude-in-chrome-troubleshooting-troubleshooting` (47 chars), well under the limit.

The plugin identity (plugin.json `name`) is unchanged at `claude-in-chrome-troubleshooting`.

## Test plan

- [x] Verified locally: after both commits + Codex restart, the load-error stops firing in `codex-tui.log`.
- [x] Local clone has both commits applied; restart of Codex TUI confirms clean skill load.


siteexperts:
Closing — keeping the fix local on my end only. No further action needed from maintainers.

PR #169, opened 2026-05-20
Fix for when CLAUDE_PLUGIN_ROOT contains spaces

VirtualTim:
My Claude is installed under `~/Library/Application Support`. Before this change I was getting this error:
```
Stop hook error: Failed with non-blocking status code: /bin/sh: ~/Library/Application: No such file or directory
```

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -365,7 +365,7 @@
     },
     {
       "name": "skill-improver",
-      "version": "1.0.1",
+      "version": "1.0.2",
       "description": "Automatically reviews and fixes Claude Code skills through iterative refinement until they meet quality standards. Requires plugin-dev plugin.",
       "author": {
         "name": "Paweł Płatek",
--- a/plugins/skill-improver/.claude-plugin/plugin.json
+++ b/plugins/skill-improver/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "skill-improver",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "Automatically reviews and fixes Claude Code skills through iterative refinement until they meet quality standards. Requires plugin-dev plugin.",
   "author": {
     "name": "Paweł Płatek",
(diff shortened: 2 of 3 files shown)


GrosQuildu (dismissed):
Nice finding. Thanks.

dguido:
## Review Summary

Thanks for the fix — this is a real bug. When `CLAUDE_PLUGIN_ROOT` resolves to a path containing spaces (e.g. `~/Library/Application Support/...`), the shell word-splits the unquoted `command` string, producing the `No such file or directory` error you hit. Quoting the path is the correct fix.

I verified the change repo-wide: `skill-improver` was the **only** hook command missing quotes — `gh-cli` and `modern-python` already quote (via `bash "${CLAUDE_PLUGIN_ROOT}/..."`), and `fp-check` uses prompt-type hooks. `stop-hook.sh` is committed executable (`100755`) with a `#!/bin/bash` shebang, so direct quoted execution works.

### Findings

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | P1 | Unquoted `${CLAUDE_PLUGIN_ROOT}` breaks hook when path contains spaces | Fixed by author (quoting) — verified correct |
| 2 | P2 | No version bump — clients only pull plugin updates when the version increases, so existing users wouldn't receive this fix | Fixed: bumped `plugin.json` + root `marketplace.json` 1.0.1 → 1.0.2 (kept in sync) |
| 3 | P3 | Other plugins use the `bash "..."` form rather than direct exec | Dismissed: direct exec is correct here (script is executable with a shebang); keeping the author's minimal diff |

### Verification

- **JSON validity**: plugin.json, marketplace.json, hooks.json all valid (`jq`)
- **Version consistency**: both manifests at 1.0.2
- **Codex skills validation**: passed (73 skills / 74 entries)

### Commit

`3e76915` — chore(skill-improver): bump version to 1.0.2 for hooks.json fix

**Determination: ready to merge.**

dguido (approved):
Fix is correct and verified repo-wide; added the version bump (1.0.2) so it ships to users. All required checks green.

PR #163, opened 2026-05-12
Update second-opinion Codex model to gpt-5.5

lightcap:
## Summary

Bumps the Codex model used by the second-opinion plugin from gpt-5.3 to gpt-5.5, and the fallback model from gpt-5.2 to gpt-5.4. Plugin version 1.6.0 → 1.7.0.

## Changes

- **SKILL.md**: Updated primary model, fallback model, and output header references
- **references/codex-invocation.md**: Updated default config, base command, fallback section, error handling table, and training note
- **plugin.json** + **marketplace.json**: Version bump to 1.7.0

## Test plan

- [x] `python3 .github/scripts/validate_codex_skills.py` passes
- [x] No remaining references to gpt-5.3-codex or gpt-5.2-codex in the plugin
- [x] Version numbers match between plugin.json and marketplace.json
- [x] Model names use `gpt-5.5` / `gpt-5.4` (no `-codex` suffix)

PR #177, opened 2026-06-02
[codex] Avoid pinning Codex model in second-opinion

### LINKED ISSUE #176: `second-opinion` skill appears to pin outdated / unavailable Codex models
Hi! Thanks for publishing this skill — I tried running the `second-opinion` skill and hit an issue with the Codex model references in `SKILL.md`.

The skill currently appears to reference `gpt-5.3-codex`, with a fallback to `gpt-5.2-codex`. When I ran it from Claude Code using my ChatGPT account, both pinned models were rejected, and the run only proceeded after falling back to the Codex CLI’s default model, which was `gpt-5.5` for my account.

### What happened

The skill first attempted to use `gpt-5.3-codex`:

```text
Auth error on gpt-5.3-codex (ChatGPT account). Falling back to gpt-5.2-codex per the skill's fallback rule:
```

Then the fallback model was also rejected:

```text
Both pinned models are rejected for this ChatGPT account. Let me check what models this account supports and retry with the default:
```

After checking the Codex CLI help/config, it found that the account’s default model was `gpt-5.5`:

```text
The account's default model is gpt-5.5. Let me run with that instead of the pinned codex models:
```

### Why this may be worth updating

OpenAI’s current Codex model documentation appears to recommend `gpt-5.5` for most Codex tasks:

https://developers.openai.com/codex/models

Because the skill pins `gpt-5.3-codex` and falls back to `gpt-5.2-codex`, it can fail for accounts that no longer have access to those pinned Codex model names, even though Codex itself is usable with the account’s current default model.

### Suggested fix

I think either of these would work:

#### Option A: update the pinned model

Update the skill to use the current recommended Codex model, for example:

```bash
codex exec -m gpt-5.5 ...
```

And update the surrounding text/output labels from `gpt-5.3-codex` to `gpt-5.5`.

#### Option B: avoid pinning a model by default

Let the Codex CLI use its configured/default model unless the user explicitly overrides it:

```bash
codex exec ...
```

Then the skill could describe the model behavior more generically, for example:

```md
Uses the configured/default Codex model. For pinned runs, pass `-m <model>` using a currently available model from the Codex model list.
```

This may be more future-proof, since the skill would not need to be updated every time the recommended Codex model changes.

### Expected behavior

The skill should run successfully for users with a working Codex CLI / ChatGPT account without requiring access to older pinned model names.

### Actual behavior

The skill attempts `gpt-5.3-codex`, then `gpt-5.2-codex`, and both can fail with auth/model availability errors before the workflow manually retries with the account’s default model.

Happy to open a PR if you prefer one of the approaches above.

FusionBrah:
## Summary

- remove the hard-coded `gpt-5.3-codex` model from the `second-opinion` Codex invocation docs
- document that `codex exec` should use the configured/default Codex model by default
- replace the outdated `gpt-5.2-codex` auth fallback with guidance for optional explicit `-m` / `--model` overrides

Closes #176

## Validation

- `git diff --check`
- `python3 .github/scripts/validate_plugin_metadata.py`
- `python3 .github/scripts/validate_codex_skills.py`

### DIFF
--- a/plugins/second-opinion/skills/second-opinion/SKILL.md
+++ b/plugins/second-opinion/skills/second-opinion/SKILL.md
@@ -172,14 +172,15 @@ structured output schema.
 
 Summary:
 - Uses `codex exec` (not `codex review`) for headless operation
-- Model: `gpt-5.3-codex`, reasoning: `xhigh`
+- Model: configured/default Codex model, reasoning: `xhigh`
 - Uses OpenAI's published code review prompt (fine-tuned into the model)
 - Diff is generated manually and piped via stdin with the prompt
 - `--output-schema` produces structured JSON findings
 - `-o` captures only the final message (no thinking/exec noise)
 - All three scopes (uncommitted, branch, commit) support project
   context and focus instructions (no limitations)
-- Falls back to `gpt-5.2-codex` on auth errors
+- For pinned runs, pass `-m <model>` or `--model <model>` with a
+  currently available Codex model
 - Output is clean JSON — parse and present findings by priority
 - Set `timeout: 600000` on the Bash call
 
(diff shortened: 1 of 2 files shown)


FusionBrah:
Just realised there is already an open PR for this exact same thing #163, closing out this one.

PR #190, opened 2026-06-24
fix(fp-check): scope hooks and make Stop path-aware

### LINKED ISSUE #143: fp-check Stop hook with matcher "*" fires on every unrelated session, causing forced LLM turns

GotenJBZ:
Fixes #143. Supersedes #188.

## Problems (in `main`)

1. `Stop` and `SubagentStop` both use `matcher: "*"`, so they run on every subagent/agent stop in any session where fp-check is installed.
2. The `Stop` prompt demands the full deep-path output (5 phases, 13 questions, full PoC) for every bug. Standard-path verifications are rejected as "incomplete".
3. Hook loops in the main agent when running fp-check using subagents/ultracode.

## Changes
1. `Stop` hook moved into skill frontmatter.  This way, the stop hooks will be executed only when the skill is loaded, avoiding the firing in unrelated sessions.
2.  Changed the `Stop` hook prompt as ordered rules: (1) skip if not fp-check (2) skip if delegated to subagents/workflow, (3) check only the path actually used (standard vs deep)
3. `SubagentStop` remains in `hooks.json`, narrowing the matcher to only support specific subagents `data-flow-analyzer|exploitability-verifier|poc-builder`

### DIFF
--- a/plugins/fp-check/hooks/hooks.json
+++ b/plugins/fp-check/hooks/hooks.json
@@ -1,21 +1,9 @@
 {
-  "description": "Enforce fp-check verification completeness — prevent premature verdicts and incomplete agent output",
+  "description": "Enforce per-worker output completeness for fp-check deep-path analysis agents. Scoped by agent type name so it fires only for fp-check's own workers.",
   "hooks": {
-    "Stop": [
-      {
-        "matcher": "*",
-        "hooks": [
-          {
-            "type": "prompt",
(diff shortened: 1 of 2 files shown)


GotenJBZ:
Adding as a reviewer both @remiforall and @computerality since they both tried to solve this exact problem. 

EDIT: I cannot do that :ops:

dguido:
Thanks for this @GotenJBZ — the diagnosis is right on all three counts, and #205 now tracks the part that isn't covered elsewhere.

Closing in favour of #188, and I want to be straight about why, because it isn't "ours over yours":

**Where #188 goes further.** Your PR narrows the `SubagentStop` matcher to `data-flow-analyzer|exploitability-verifier|poc-builder`. #188 deletes `plugins/fp-check/hooks/hooks.json` entirely and moves each completeness check into the frontmatter of the agent it belongs to. Same insight as yours — frontmatter scoping — pushed one step further: with no hooks file there is no global matcher left to get wrong later.

**Where your PR goes further.** #188 moves the `Stop` prompt but does not change it, so your problem 2 is real and still open: a standard-path verification is still measured against the deep path and reported incomplete. That is #205, with your rule ordering quoted and credited.

**Why #188 goes first.** It is CI-validated and bumps 1.0.2 → 1.0.3 in both `plugin.json` and `marketplace.json`. #190 has run only the CLA check (fork PRs need maintainer approval to run workflows) and carries no version bump, so clients would not receive it even if merged.

If you want to take #205, it's yours — and it's the more interesting half of the problem.

GotenJBZ:
> Closing in favour of https://github.com/trailofbits/skills/pull/188, and I want to be straight about why, because it isn't "ours over yours":

Sounds good to me, but imho the approach in that PR has a big problem, as mentioned by @computerality [here](https://github.com/trailofbits/skills/pull/188#issuecomment-4764445803), Claude Code doesn't load hooks in the frontmatter coming from a plugin subagent.

Until Anthropic changes the policy, users will need to copy-and-paste the plugin without using the marketplace to correctly use the skill

> so your problem 2 is real and still open

That's true tbh, I can open a pr on top of #188 for the second one, just wanted to cross-check for the other issue before doing that

PR #173, opened 2026-05-27
Remove legacy codex compatiblity scripts/shims.

Ninja3047:
Codex supports claude plugins so this shouldn't be necessary. Add a script to test the plugin loadablility in both claude and codex

PR #193, opened 2026-06-25
feat: add OpenCode native plugin support

myphsto:
## Summary

Add native OpenCode plugin support via `.opencode/plugins/trailofbits-skills.js` which automatically discovers skills from `plugins/*/skills/` directories.

## Changes

- **`.opencode/plugins/trailofbits-skills.js`** — ES module plugin that auto-discovers all `plugins/*/skills/` directories and injects them into OpenCode's config
- **`package.json`** — root-level `{"type": "module"}` required by Node.js v26 for ES module support
- **`.opencode/.gitignore`** — ignores `node_modules/`
- **`.github/scripts/check_opencode_loadability.py`** — end-to-end CI validation: runs `opencode debug skill`, verifies all skills load with valid frontmatter
- **`.github/workflows/validate.yml`** — installs OpenCode CLI, runs loadability check
- **`README.md`** — adds OpenCode installation section
- **`AGENTS.md`** — updates cross-platform compatibility section, adds OpenCode to PR checklist

## Testing

- All 74 Trail of Bits skills load through OpenCode
- Python script passes ruff check + format
- End-to-end `opencode debug skill` validation passes
- All existing Claude Code and Codex loadability checks continue to pass

## Notes

- This is purely additive — no changes to existing Claude Code or Codex behavior
- The `.opencode/` directory is inert for all other platforms
- OpenCode auto-discovers skills, so no per-plugin metadata is needed

### DIFF
--- a/.github/scripts/check_opencode_loadability.py
+++ b/.github/scripts/check_opencode_loadability.py
@@ -1,0 +1,208 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.11"
+# dependencies = []
+# ///
+"""Check that this marketplace's skills load through OpenCode.
+
+Runs `opencode debug skill` from the repo root and verifies that every
+expected skill is present and has valid frontmatter.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+from typing import Any
+
+
+def discover_expected_skills(repo: Path) -> dict[str, Path]:
+    """Walk plugins/*/skills/*/SKILL.md and return {skill_name: path}."""
+    skills: dict[str, Path] = {}
+    plugins_dir = repo / "plugins"
+
+    if not plugins_dir.is_dir():
+        return skills
+
+    for plugin_dir in sorted(plugins_dir.iterdir()):
+        if not plugin_dir.is_dir():
+            continue
+
+        skills_dir = plugin_dir / "skills"
+        if not skills_dir.is_dir():
+            continue
+
+        for skill_dir in sorted(skills_dir.iterdir()):
+            skill_md = skill_dir / "SKILL.md"
+            if not skill_md.is_file():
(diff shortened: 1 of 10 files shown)


myphsto:
Added this on my fork figured if anyone is interested.

dguido:
Thanks for building this, @myphsto, and apologies for the wait. We're declining it — but the reason is policy, not quality, and it's worth being specific so the decision is reviewable.

**The rule.** AGENTS.md treats Claude plugin marketplace metadata as the single canonical source, and explicitly bans per-runtime sidecars (`.codex/`, `.agents/plugins/marketplace.json`, `plugins/*/.codex-plugin/`). Codex support works *because* it reads `.claude-plugin/marketplace.json` directly rather than needing its own copy. We removed the last Codex sidecars in f09e5c7 (#173) after they drifted out of sync with the canonical metadata, and today I removed two more that had crept back into #148 and #142.

**What this PR asks for.** A `.opencode/` sidecar with its own `package.json` and lockfile, a **root-level `package.json` setting `"type": "module"` for the whole repository**, a CI job installing the OpenCode CLI as a build dependency, and an AGENTS.md edit adding OpenCode to the contribution checklist. That reverses the stance for a third runtime, and it lands the cost on every contributor: a repo-wide module type, a new CI dependency, and one more thing to keep in sync on every plugin PR.

**What would change our minds:**

1. **OpenCode reads the Claude marketplace metadata directly,** the way Codex does. Then there is nothing to add — it just works, and we'd happily document it.
2. **Someone here actually uses OpenCode.** Right now nobody does, so we'd be maintaining a compatibility layer with no one to notice when it breaks — which is precisely how the Codex sidecars rotted.

If (1) is achievable upstream, that's the version we'd take, and I'd be glad to look at it. As a note for anyone finding this later: the sidecar ban is now enforced in CI rather than left to review, and `.opencode/` is included in that check — not to pre-judge this decision, but so the repo state and the written rule can't drift apart again.

PR #196, opened 2026-07-06
fix(modern-python): suggest exact `uv run python` so shim advice works outside projects

Ninja3047:
## Bug

As reported in #195: the `python`/`python3` shim's suggested command can be a dead end. The shim echoes back whichever name was invoked (`uv run $cmd ...`), and for `python3` that advice fails on any machine with **no uv-managed interpreters** (fresh CI runners, devcontainers, stock Ubuntu/Debian): outside a project, uv resolves the `python3` *command* through an ordinary PATH lookup, which hits the shim again and errors with the same suggestion — an infinite advice loop for the agent.

Reproduced on a stock `ghcr.io/astral-sh/uv:debian` container (apt `python3`, zero managed Pythons, no env overrides, no project):

```console
$ uv run python3 script.py
ERROR: Use `uv run python3 script.py` instead of `python3 script.py`
```

## Fix

uv special-cases the exact command name `python` (`target.eq_ignore_ascii_case("python")` → `RunCommand::Python` in `crates/uv/src/commands/project/run.rs`, present since at least uv 0.4.0): `uv run python ...` executes uv's resolved interpreter directly with **no PATH lookup**, so it can never resolve back to the shim.

The shim now always suggests the exact-`python` spelling — `uv run python ...` / `uv run python -m ...` — regardless of whether `python` or `python3` was intercepted (the "instead of" half still quotes what was actually run). Verified on the same stock Debian container and on macOS that the suggested form succeeds for script, `-c`, `-m module`, and REPL invocations, inside and outside projects.

This is an alternative to the passthrough approach in #195: the shim stays a pure print-and-exit blocker (no `UV` env detection, PATH walking, or exec-loop guards), keeps working when `UV` leaks in from ancestors like `uv run claude` or Agent SDK apps launched via `uv run`, and lowers the uv floor from 0.6.0 to 0.4.0. Known non-goal: nested tooling under `uv run` that hardcodes `python3` (e.g. a Makefile) still gets blocked with advice it can't follow mechanically — acceptable for an agent-facing nudge, and #195's passthrough remains a possible follow-up if that case matters.

Also updates the README/`setup-shims.sh` rationale (the old "uv run prepends its venv bin/" claim only holds inside a project or with managed interpreters installed).

## Tests

- `python-shim.bats`: two new tests assert the suggestion uses exact `uv run python` (never `python3`) when invoked via the `python3` symlink, for both script and `-m` forms; the existing symlink test now checks the "instead of" side. All 10 pass (bats 1.12), full shims suite 35/35 green.
- `shellcheck -x` (0.11.0) and `shfmt -i 2 -ci` (3.13.1) clean.
- `check_claude_loadability.py` / `check_codex_loadability.py` pass.
- Version bumped 1.5.2 → 1.5.3 in `plugin.json` and root `marketplace.json`.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -262,7 +262,7 @@
     },
     {
       "name": "modern-python",
-      "version": "1.5.2",
+      "version": "1.5.3",
       "description": "Modern Python best practices. Use when creating new Python projects, and writing Python scripts, or migrating existing projects from legacy tools.",
       "author": {
         "name": "William Tan",
--- a/plugins/modern-python/.claude-plugin/plugin.json
+++ b/plugins/modern-python/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "modern-python",
-  "version": "1.5.2",
+  "version": "1.5.3",
   "description": "Modern Python best practices. Use when creating new Python projects, and writing Python scripts, or migrating existing projects from legacy tools.",
   "author": {
     "name": "William Tan",
--- a/plugins/modern-python/README.md
+++ b/plugins/modern-python/README.md
@@ -37,7 +37,7 @@ Modern Python tooling and best practices using uv, ruff, ty, and pytest. Based o
 
 ## Hook: Legacy Command Interception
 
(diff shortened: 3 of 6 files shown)


Ninja3047:
@claude review once

tob-scott-a (approved):

kz-tob (approved):

PR #201, opened 2026-07-20
Add vulnerability triage brocards skill

DarkaMaul:


dguido (approved):
Approving on the contribution checklist and CI, reviewed with Claude Code.

Verified:
- Frontmatter parses; `name` is kebab-case and matches the directory
- Description is third person and carries concrete trigger phrases
- All three required sections present: `## When to Use`, `## When NOT to Use`, `## Rationalizations to Reject`
- Version parity 0.1.0 between `plugin.json` and `marketplace.json`
- Registered in marketplace.json, root README table, and CODEOWNERS
- SKILL.md is 208 lines; one level of references, no chains
- No hardcoded absolute paths; `skills/` at plugin root
- All 7 checks green after updating the branch

Not verified: the substance of the 7 brocards as triage guidance. That rests on @DarkaMaul and the linked source.

PR #192, opened 2026-06-25
Add github-triage plugin

ESultanik:
## Summary

Adds a new `github-triage` plugin: a single, explicitly-invoked skill (`/github-triage`) that triages a repository's open GitHub issues and pull requests via the `gh` CLI.

**Repository selection** — uses the sole GitHub remote by default; prompts when PWD is not a git repo, has no GitHub remote, or has several.

**Pull requests (optional, run first so merges feed issue resolution):**
- Incrementally merge allowlisted bot PRs (Dependabot/Renovate) and maintainer-approved PRs — one at a time, oldest first, re-checking mergeability/CI before each and confirming it landed.
- Spawn one read-only review subagent per never-reviewed (human-authored) PR; each review is saved to `github-pr-<number>-review.md` locally and never posted.

**Issues:**
- Close already-resolved issues with a comment citing the resolving PR/commit.
- Cross-link issues and their pending fix PRs (non-destructive comment; body-edit for auto-close is opt-in).
- Assign **local-only** priority (Critical/High/Medium/Low) and change-size (`size/XS`–`XXL`, Kubernetes/Prow thresholds) to everything outstanding — never posted to GitHub. Large sets (>32) can be saved to Markdown.

All GitHub writes — merges included — are gated behind a single review-and-iterate approval; nothing runs autonomously.

## Validation

- Verified PR-readiness logic against live `gh --json` field semantics: `mergeStateStatus == CLEAN` as the merge gate (`MERGEABLE` alone is insufficient; `UNKNOWN` is never-merge), per-node `statusCheckRollup` (`CheckRun` vs `StatusContext`), `author.is_bot` + a normalized allowlist, and `latestReviews` state + `authorAssociation` rather than the branch-protection-driven `reviewDecision`.
- Ran an end-to-end **dry run against trailofbits/graphtage** (19 issues, 7 PRs, no writes), which caught and fixed four real bugs: `gh repo view` is positional (not `-R`); `NEUTRAL`/`SKIPPED` checks must not block merge; not-ready bot PRs belong in "Needs work", not the review bucket; and the bot allowlist must match `gh`'s `app/dependabot` login form.
- All repo CI gates pass: `check_claude_loadability.py`, `check_codex_loadability.py`, `validate_plugin_metadata.py`. SKILL.md is 410 lines (<500). Registered in the marketplace, README, and CODEOWNERS.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #206, opened 2026-07-29
Move the contribution checklist into machinery

dguido:
AGENTS.md documented roughly 53 rules and CI machine-enforced 6 of them. This closes the gap for the ones a machine can decide, deletes the checklist for those, and keeps prose only for what a tool genuinely cannot judge.

Two things drove the shape. Current model guidance says verification scaffolding in prompts degrades output rather than improving it, and that removing it costs no capability — so the answer to an unenforced rule is machinery, not a louder checklist. And the defect class that actually reaches main here is a checker reporting success while inspecting nothing, so everything is built around one rule:

> **A checker that inspects zero items must fail, not pass.**

## What is enforced now

Every new error-level check passes on all 42 plugins today, so **nothing here blocks anyone on day one**. Regression guards, plus one real gap:

| Check | Why |
|---|---|
| Agent files use `tools:`, skills use `allowed-tools:` | Keys are inverted between file types; the loader silently ignores the wrong one, so the restriction does not apply |
| `subagent_type` namespaced `<plugin>:<agent>` | A bare name is unregistered and the dispatch fails at runtime |
| **Version increment vs. base branch** | The real gap. `debfb29` shipped an `allowed-tools` fix across 25 plugins and bumped none, so no installed user received it. Scoped to plugins the branch touched; `no-version-bump` label skips it |
| Forbidden sidecars | AGENTS.md banned these with zero enforcement; two open PRs had them |
| kebab-case, ≤64 chars, plugin `README.md`, semver | Previously human-review only |

Warning-level and non-blocking: required SKILL.md sections, the 500-line limit, unresolved references. **55 warnings across 42 plugins**, and the distribution matters more than the total:

| Plugin | Warnings |
|---|---|
| `testing-handbook-skills` | 29 |
| `building-secure-contracts` | 20 |
| everything else combined | 6 |

Two plugins are 49 of 55. That is a conversation with those owners, not a sweep, and nothing here forces it.

## The anti-vacuity guards

1. **`--self-test`** builds fixtures and asserts every checker rejects a known-bad one. Runs first in CI. 29 assertions.
2. **Zero-reference guard**: a full scan resolving zero references exits 1. Computed from scan coverage, not from whether `CHANGED_FILES` was empty — every plugin PR touches `marketplace.json`, which sweeps all plugins, so keying off changed files would disarm it on essentially every real PR.
3. **`SELF_TEST_MINIMUM`**: the self-test fails below 20 assertions, because the self-test is itself a checker.

Plus a Makefile↔CI ruff parity assertion. Both it and the version-increment check were verified by deliberately breaking them and confirming red.

## CI checks that could pass or fail without inspecting anything

Four instances of the same bug, found while building the above:

- **`python-tests` ran `python3 <file>` per file** — a file with no `__main__` block exits 0 having run nothing. Now pytest per directory: **278 tests run** where the count was previously unknown.
- **`bats` used `--no-run-if-empty`** — broken discovery was a silent pass.
- **The frontmatter walk** printed `All 0 SKILL.md files have valid frontmatter` if it broke.
- **`zeroize-audit`'s shell suites** matched no CI glob and had never run.

And the inverse: `check_codex_loadability.py` raced the codex server on tempdir cleanup and failed the job *after* every check had passed (seen on #148 today).

## Automated PR review

Two tiers: fast on every push, deep on a `deep-review` label. **Correction to an earlier version of this description**: I originally wrote that this was inert for want of an `ANTHROPIC_API_KEY`, based on `gh secret list` returning empty. That only lists *repository* secrets — an org-level key is visible here, so **the fast tier is live on merge**. It has already reviewed this PR twice.

The prompt **forbids pre-filtering**: "only report high-severity issues" is followed literally, so the model investigates fully, finds the bugs, then declines to report them. Recall appears to collapse and it looks like a capability regression when it is a prompt bug.

**Fork PRs get no automated review.** A `pull_request_target` variant was written and deleted: it checked out the fork tree and ran this repo's review script *from that tree*, so the script itself — and any `CLAUDE.md` or `.claude/hooks` beside it — was fork-authored and executed with the org key in the environment. The SHA pin and tool allowlist did not touch that path. Doing it safely means base in the workspace and the fork commit in a subdirectory; that is deliberate work, not a footnote here.

## This PR was reviewed by the thing it adds, twice, and both passes found real bugs

Worth recording, because it is the best evidence the prompt works:

- **Pass 1** found that the fork workflow executed fork-controlled code (the P1 above), that the prompts asked for `--body-file <file>` with no `Write` tool available, that `no-version-bump` was named everywhere and implemented nowhere, and that `make check` was not the CI superset it claimed.
- **Pass 2** found that the review script **could not distinguish "reviewed clean" from "never posted"** — it ended on `claude --print` and exited 0 regardless, so a silent run produced a green check. Defect class 1, in the script that ships the prompt warning about defect class 1. It also caught `_check_ruff_parity` returning a vacuous pass when its inputs were missing, inside the anti-vacuity harness.

Both are fixed. Two findings were accepted without change, with reasoning in the commit message: the reference resolver's looseness, and the subagent check not knowing other plugins' agents.

## Also

- **pre-commit**: the validators were CI-only; AGENTS.md asked people to remember them. Adds actionlint, zizmor, check-toml, check-merge-conflict, detect-private-key.
- **Markdown linting deliberately omitted**, measurement recorded in the config: ~12,400 violations (7,282 MD013 alone). It would land permanently red or gutted to the point of checking nothing.
- **Dependabot**: the `pip` entry at `/` had no manifest to resolve, so it reported nothing indefinitely; two script dirs with real dependencies were uncovered.
- **AGENTS.md**: the frontmatter example showed `allowed-tools` as a YAML list. Every SKILL.md uses the space-delimited form — the doc was the outlier and it produced a false finding while reviewing #192.
- **#207** filed separately: the `modern-python` shims reject `uv pip install` (breaking prek hook installation) and `python3 -` (breaking zeroize-audit's suite). That is why `make shell-suites` exists but is not part of `make check`.


dguido:
@claude review

dguido:
@claude review

PR #175, opened 2026-05-30
feat(codex): add UI metadata for skills

2023Anita:
## Summary

- add a shared Codex UI mark for Trail of Bits skills under `.codex/assets/`
- add `agents/openai.yaml` UI metadata for every plugin skill and the Codex-only `gh-cli` entry
- keep the metadata limited to `interface` fields: `icon_small`, `icon_large`, and `brand_color`

## Motivation

Closes #174.

This makes the Trail of Bits skills easier to identify in Codex UI surfaces without changing the skills' instructions, routing, plugin manifests, or runtime behavior.

## Verification

```bash
python3 .github/scripts/validate_codex_skills.py
python3 .github/scripts/validate_plugin_metadata.py
git diff --check
```

I also checked that all `agents/openai.yaml` icon paths resolve locally.

PR #214, opened 2026-07-30
Remove ask-questions-if-underspecified

kevin-valerio:
Removes the `ask-questions-if-underspecified` plugin.

The skill content is untouched since 2026-01-29. The only commits to the directory
since were repo-wide sweeps (formatting in #98, codex UI metadata in #175), not
maintenance.

It also spends a lot of context on something you can get from one English sentence
to Claude. Asking clarifying questions when a request is ambiguous is native Claude
Code behaviour now, through `AskUserQuestion`, so an 86-line skill that mostly
explains how to format multiple-choice options is paying context for a built-in.

Two knock-on edits, both forced by the removal:

`AGENTS.md` used this plugin as its **Basic** reference example. That now points at
`git-cleanup`, which is a better fit anyway: five files, one self-contained
`SKILL.md`, nothing to chase.

`let-fate-decide` named the skill in its description and in its When NOT to Use
list. Both reworded to plain "ask clarifying questions". Its description feeds skill
triggering, so it gets a patch bump to 1.2.2 in `plugin.json` and `marketplace.json`.

Registry entries in `marketplace.json`, `README.md` and `CODEOWNERS` are gone. The
validator reports no errors with 359 references resolved, and grep finds zero
remaining mentions of the removed name.


kz-tob (dismissed):
Looks good

tob-scott-a (approved):

kz-tob (approved):

Issue #12, opened 2026-01-16
docs: Add Gemini CLI compatibility instructions


dguido:
## Summary

Research into Gemini CLI compatibility for these skills.

### Skills Compatibility

Gemini CLI v0.23.0+ has experimental Agent Skills support. Our skills work via the
[gemini-cli-skillz](https://github.com/intellectronica/gemini-cli-skillz) bridge extension
with **no modifications needed**.

**Key differences:**
- Gemini only supports `name` and `description` in frontmatter (no `allowed-tools`)
- No plugin.json wrapper
- Tool restrictions are CLI-level, not per-skill

**User setup:**
```bash
# Enable experimental skills
gemini  # then /settings → search "Skills" → enable

# Install skillz bridge
gemini extensions add gemini-cli-skillz
```

### Security Extension (Complementary)

Gemini CLI also has a [security extension](https://github.com/gemini-cli-extensions/security)
that provides vulnerability scanning:

- `/security:analyze` - analyze code changes for vulnerabilities
- `/security:scan-deps` - OSV-Scanner integration for dependency vulnerabilities
- 90% precision, 93% recall on OpenSSF CVE Benchmark
- GitHub Actions integration for PR reviews

This could complement our security audit skills.

## Proposed Changes

1. Add `docs/gemini-cli.md` with installation instructions
2. Mention security extension as complementary tool
3. Update README.md to note Gemini CLI compatibility

## Sources

- [Gemini CLI Skills Docs](https://geminicli.com/docs/cli/skills/)
- [gemini-cli-skillz Extension](https://github.com/intellectronica/gemini-cli-skillz)
- [Security Extension](https://github.com/gemini-cli-extensions/security)

PR #217, opened 2026-07-30
git-cleanup: convert the skill to a command driving a dynamic workflow

frabert:
Replaces the prose `SKILL.md` with a `/git-cleanup` slash command plus a [dynamic workflow](https://claude.com/blog/a-harness-for-every-task-dynamic-workflows-in-claude-code) — a JavaScript orchestration script that fans branch analysis out across subagents.

## What changed

| Path | Role |
|------|------|
| `commands/git-cleanup.md` | The `/git-cleanup` entry point: both gates, the confirmations, and every deletion |
| `workflows/analyze-branches.js` | The dynamic workflow. Read-only — it never deletes |
| `references/merge-evidence.md` | What counts as proof a branch is merged, read by the agents and the fallback path |
| `tests/analyze-branches.test.mjs` | 37 assertions over the deterministic core, every agent stubbed |

`skills/git-cleanup/` is deleted. Version 1.0.1 → 2.0.0 in both manifests — removing the skill is a capability removal, not a feature.

## Why this shape

Only the genuinely uncertain branches get an agent:

1. **Survey** — one agent inventories branches, worktrees, tracking state, merge history.
2. **Triage** — plain JavaScript. Protected-branch filtering, merged branches, unpushed work, and branches level with a live remote are decided from git's own output. Nothing spawns for a question `git branch --merged` already answered.
3. **Investigate** — batched agents hunt merge evidence for the remainder. Related branches travel together, since supersession is only visible when one agent sees the whole cluster.
4. **Refute** — every squash-merged/superseded candidate goes to a skeptic told to find a commit that is *not* in the default branch.

The fleet is bounded at eleven agents; past five batches the batches grow rather than the count.

**Nothing destructive moved into the workflow.** Subagents run in the background with no way to reach the user, so both gates and every `git branch -d/-D` and `git worktree remove` stay in the main session. Every agent prompt carries a read-only constraint.

**Uncertainty resolves toward keeping a branch.** A refutation missing its `refuted` field, duplicate refutations, a dead agent, and a missing verdict all downgrade to needs-review. A wrong keep costs another look at a branch list; a wrong delete costs work that exists nowhere else.

## Review notes

- **A multi-agent review of this branch found 15 defects, all fixed here** — three independent fail-open paths in the refutation gate, a protected-branch filter that never excluded the repo's actual default branch (so a repo on `trunk` was offered `git branch -d trunk`), a worktree marked stale without consulting its dirty flag, and branch names reaching Bash unquoted (git refnames forbid spaces but permit `$` and backticks, so `$(id)` is a legal branch name). Each was reproduced by execution before and after the fix.
- **New `js-tests` make target and CI job.** Both carry the same zero-discovery guard as `python-tests` — an empty glob fails rather than reporting a pass. Verified the guard fires. The earlier version of this branch shipped the suite with nothing running it, and a mutation that silently dropped unverified candidates passed 22/22; it now fails.
- **The plugin deliberately no longer ships a skill.** The point of the change is that workflow-shaped work should be a workflow rather than a Skill, so the skill is not coming back as a Codex anchor. Two consequences, both accepted: it loses `agents/openai.yaml` and the brand mark (that presentation metadata only attaches to skills here), and `check_codex_loadability.py` — which counts `skills/**/SKILL.md` — now passes vacuously at `0 == 0` for this plugin. Nothing in CI verifies the plugin's only entry point loads under either CLI. That gap is in the checkers, not in this plugin, and is worth its own issue.
- **`.js`/`.mjs` remains unlinted and unformatted.** No formatter in this repo covers it. Left out of scope; wiring one up is a repo-wide decision.

## Testing

`make self-test lint python-tests js-tests validate` is green — validator reports no errors, 361 references resolved, 37/37 JS assertions.


frabert:
Thanks — worked through all of it in b6aaee2. Every P2 and P3 was real and is fixed.

**P2 — deletion order.** Right, and the README transcript was reporting an outcome that command sequence cannot produce. Gate 2, phase 3, and the transcript now put `git worktree remove` before the branch delete, ordered from the `worktreePath` each candidate already carries. Phase 3 also now says what to do if a delete still fails with "used by worktree at …": report it, don't remove a worktree that was never in the gate-2 confirmation.

**P2 — unbounded unit size.** Fixed with `MAX_BRANCHES_PER_UNIT = 10`; oversized clusters split, each slice keeping the full context list so the superseding sibling stays visible to every slice. Verified with 40 `dependabot/npm_and_yarn/*` branches: 1 unit / 1 investigator before, 4 units / 2 after.

**P2 — refuter asked the wrong question.** The clearest of the three. The refuter now tests each claim against what the claim named — the default branch for a PR or commit, the superseding branch for a supersession — instead of always the default branch. `REFUTE_SCHEMA`'s description of `refuted` said the same wrong thing and is updated too.

**P3 — `fetch --prune` vs `READ_ONLY`.** This was the one I'd rank highest, because the failure is silent and total: an agent resolving the contradiction in favour of the hard constraint sees no `[gone]` branches and the tool reports a clean repo. `fetch --prune` is now explicitly permitted in the constraint, with the reason.

**P3 — suite printed "passed" on a failing run**, and **P3 — `js-tests` guarded discovery but not execution.** Both were my guard being half a guard. The suite now tracks failures separately and prints `FAIL n of m assertions failed`; the runner requires each suite to emit `<n> assertions passed` with n > 0, so `node <file>` on a file that asserted nothing fails instead of counting as a suite. Verified both by mutation.

**P3 — `pluginDir` silent failure.** The inline standard is now unconditional and the reference is additive, so a failed Read costs detail rather than the whole evidence standard.

**Version.** Agreed, and taken to 2.0.0.

Also fixed from the P4 list: `node` documented as a `make check` prerequisite, `unpushedCommits` specified for a gone upstream (-1, not 0 — reporting 0 would assert something unmeasured), the `mergeLog` 40-entry window documented as a window, `meta.whenToUse` no longer naming the deleted skill, and the `|| echo main` comment that described code the fix had already removed.

**Two things deliberately not changed:**

- **The Codex loadability gap stands, and the skill is not coming back.** The intent of this PR is that workflow-shaped work should be a workflow rather than a Skill, so restoring a skill purely as a Codex anchor would undo the point. You're right that this is more than a maintainer's preference: with `check_codex_loadability.py` counting only `skills/**/SKILL.md` and `check_claude_loadability.py` checking nothing about `commands/`, no CI check verifies this plugin's only entry point loads under either CLI, and the codex check now passes at `0 == 0`. That is a gap in the checkers rather than in this plugin — it applies to any commands-only plugin — and I'd rather fix it there, in its own PR, than reshape this one around it.
- **`EXPECTED_ASSERTIONS` stays hand-maintained.** It reads like a regression when you forget to bump it, but that constant is the mechanism that makes the counter meaningful; deriving it from the run would make it unfalsifiable.

One note on history: the branch was rebased onto #216 and force-pushed while I was working. I replayed my commit onto the new base with `--onto` after confirming the rebased commits were content-identical to mine — nothing of that push was discarded.

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,547 @@
...
+for (const b of branches) {
+  if (b.merged) {
+    // Reported by the survey agent as listed in `git branch --merged`. This is the one
+    // delete category that skips investigation and refutation, so it is NOT independently
+    // verified here — `git branch -d` re-derives the merge and refuses if it is wrong,
+    // which is why this category is pinned to `-d` and never to `-D`.
+    settled.push({
+      ...b,
+      category: 'SAFE_TO_DELETE',
+      evidence: `reported merged into ${survey.defaultBranch}; git branch -d re-checks`,
```
hbrodin: **Blocker — the `-d` backstop checks a different property than this comment claims.**

`SAFE_TO_DELETE` is the one delete category that skips both investigation and refutation, on the grounds that `git branch -d` re-derives the merge. It doesn't. `-d` accepts a branch merged into **HEAD** *or* into its **own upstream** — neither of which is "merged into `defaultBranch`".

Reproduced on git 2.51.0:

- Merged to upstream, not to main: `git branch -d synced` -> `Deleted branch synced`, exit 0. Is that commit reachable from `main`? **No.**
- With `HEAD=integration`, `git branch --merged main` does *not* list `sub`, yet `git branch -d sub` -> `Deleted branch sub`, exit 0.

So it errs toward **accepting**, not refusing. Damage is capped in both cases (the work survives on the remote or in HEAD), but the safety argument as written doesn't hold, and this is the only delete category with nothing behind it.

Two ways out: route `SAFE_TO_DELETE` through the refutation pass like the other two categories, or have the survey report the merge commit alongside the `merged` flag so the claim is checkable rather than asserted.

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,547 @@
...
+    { title: 'Survey', detail: 'one agent inventories branches, worktrees, and merge history' },
+    { title: 'Investigate', detail: 'batched agents hunt merge evidence for the ambiguous branches' },
+    { title: 'Refute', detail: 'batched skeptics try to prove each delete candidate still holds unique work' },
+  ],
+}
+
+// Branches that must never be analyzed, recommended, or deleted. Filtered here in
+// JavaScript rather than in an agent prompt: an agent can be talked out of a rule,
+// a regex cannot.
+const PROTECTED = /^(main|master|develop|release\/.*)$/
```
hbrodin: **Blocker — `staging`, `production`, `dev`, and `hotfix/*` all reach the delete list.**

Ran the workflow with stubbed agents against a survey containing those four names:

```
merged path      -> [["staging","git branch -d"],["production","git branch -d"],
                     ["dev","git branch -d"],["hotfix/urgent","git branch -d"]]
remote-gone path -> [["staging","git branch -D"],["dev","git branch -D"],
                     ["production","git branch -D"],["hotfix/urgent","git branch -D"]]
remote-gone needsReview -> []
```

The remote-gone path yields **force-delete** for all four, with an empty `needsReview`.

The `defaultBranch`/`currentBranch` additions genuinely fixed the `trunk` case — that was a good catch. But the literal list here is thinner than the prose around it suggests, and these are exactly the long-lived names whose remote gets deleted during a branch-protection change or a repo migration.

plugins/git-cleanup/references/merge-evidence.md
```diff
@@ -0,0 +1,66 @@
...
+| `SQUASH_MERGED` | A named commit or PR number in the default branch whose diff carries this branch's changes. | `git branch -D` |
+| `SUPERSEDED` | Either a named PR that merged the work, or a named newer branch that contains every commit of the older one. | `git branch -D` |
+| `REMOTE_GONE` | The upstream is `[gone]` and the work was not located in the default branch. | none — review |
+| `UNPUSHED_WORK` | `git log <upstream>..<branch>` is non-empty. | none — keep |
+| `LOCAL_WORK` | Untracked branch holding commits that exist nowhere else. | none — keep |
+| `SYNCED_WITH_REMOTE` | Level with a live upstream. Nothing local to clean. | none — keep |
+
+## Commands that produce the evidence
+
+Branch names can contain characters that break shell expansion. Always quote `"$branch"`.
```
hbrodin: The main session is hardened correctly for this — `commands/git-cleanup.md:186` demands single quotes and gives the `'\''` escape, with the right reasoning. The two files the **subagents** read are not.

This line says *always quote `"$branch"`* — double quotes, under which `$(...)` still substitutes — and the survey prompt at `workflows/analyze-branches.js:211` says only "always quote them" with no polarity. Neither is operating on a shell variable in practice; the agents paste literal branch names into commands.

Confirmed legal branch names:

```
LEGAL:    evil$(id)
LEGAL:    evil`id`
LEGAL:    has'quote
LEGAL:    a;b
LEGAL:    a|b
rejected: a b      <- only the space is refused
```

`has'quote` being legal also means the `'\''` escape in the command file is load-bearing rather than theoretical. Suggest mirroring the single-quote rule into this file and into the survey prompt, since these are the copies the agents actually follow.

plugins/git-cleanup/commands/git-cleanup.md
```diff
@@ -0,0 +1,251 @@
...
+git branch --merged "$default_branch"
+git worktree list --porcelain
+git log --oneline "$default_branch" | grep -iE "#[0-9]+" | head -40
+```
+
+## Phase 2: Check the Workflow's Work
+
+The workflow reports evidence so you can audit it, not so you can forward it unread. Before building the gate-1 table:
+
+1. **Every `deleteCandidate` names specific evidence** — a PR number, a commit sha, or a superseding branch. "Similar name", "looks stale", or an empty evidence string is not a delete recommendation. Move it to needs-review.
```
hbrodin: This audit rule rejects the workflow's own output.

Every `SAFE_TO_DELETE` candidate carries exactly this evidence string, from `workflows/analyze-branches.js:282`:

```
"reported merged into main; git branch -d re-checks"
```

No PR number, no commit sha, no superseding branch. Applied literally, this rule moves **every** git-proven merged branch to needs-review — the one category where the evidence is strongest.

Either exempt `SAFE_TO_DELETE` here explicitly, or have the workflow name the merge commit in that evidence string so it satisfies the rule on its own terms.

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,547 @@
...
+            type: 'integer',
+            description:
+              'count of git log <upstream>..<branch>. Use -1 when the branch has no upstream ' +
+              'configured. Use -1 ALSO when the upstream is configured but gone from the ' +
+              'remote, since that ref cannot be compared against — do not report 0, which ' +
+              'would assert the branch holds nothing unpushed when you did not measure it.',
+          },
+          uniqueCommits: { type: 'integer', description: 'count of git log <default>..<branch>' },
+          lastCommit: { type: 'string', description: 'short sha and subject' },
+          worktreePath: { type: 'string', description: 'checkout path, or "" when none' },
```
hbrodin, line 128: `worktreePath` is optional here — it's absent from the `required` list above — but load-bearing downstream: `commands/git-cleanup.md:157` tells the model to order `git worktree remove` before the branch delete using this field.

A schema-conformant survey that omits it produces:

```
deleteCandidates: [{"branch":"feature/auth", ..., "worktreePath":""}]
worktrees:        [{"path":"/wt/auth","branch":"feature/auth","stale":true}]
```

The ordering is then silently skipped and the branch delete fails with `used by worktree at ...`. The command handles that failure gracefully, so this degrades rather than breaks — but it degrades for exactly the case the analysis flagged.

The authoritative source is the **required** `worktrees[]` array, which `report()` already joins against at line 541. Deriving the ordering there removes the dependency on an optional field.

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,547 @@
...
+
+// Clustered branches travel together: supersession is only visible when one agent sees
+// the whole cluster at once. `context` carries the siblings that triage already settled —
+// usually the newer, still-tracked branch that superseded the others. Withholding them
+// would ask the agent to name the superseding branch while hiding it. They are shown to
+// the agent but no verdict is accepted for them.
+const pending = new Set(ambiguous.map((b) => b.name))
+const units = clusters.flatMap((c) => {
+  const decide = c.filter((b) => pending.has(b.name))
+  const context = c.filter((b) => !pending.has(b.name))
```
hbrodin: `MAX_BRANCHES_PER_UNIT` caps `decide` at 10, but `context` is uncapped and gets replicated into **every** slice below (line 322).

Measured with 300 settled siblings and 10 ambiguous branches in one cluster:

```
decideRefsPerPrompt:  [10]
contextRefsPerPrompt: [600]
promptChars:          [40872]
```

~41 KB of prompt to decide ten branches, ~98% of it context.

For balance: the `decide` side is fine. I measured 150 branches -> 5 investigators -> **30 branches per prompt**, so that cap does what the comment at lines 28-33 intends. It's only the context list that grows without bound.

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,547 @@
...
+  'or the superseding branch that carries its work. A shared name prefix, the branch being',
+  'old, and a `[gone]` upstream are all non-evidence.',
+  pluginDir
+    ? `The full standard, with the commands that produce each kind of evidence, is in ${pluginDir}/references/merge-evidence.md — read it. If that path does not resolve, say so in your evidence field and apply the standard above.`
+    : '',
+]
+  .filter(Boolean)
+  .join('\n')
+
+const READ_ONLY = [
```
hbrodin, line 82: `READ_ONLY` is a prose constraint, and none of the three `agent()` calls (lines 215, 380, 411) pass `agentType` — so nothing restricts these subagents' tools, and they must hold Bash to run git at all.

Three attacker-influenceable strings are interpolated verbatim into those prompts:

| Source | Site |
|---|---|
| `mergeLog` (default-branch commit subjects) | lines 341-345 |
| `lastCommit` subject + branch names | lines 361-362 |
| investigator `evidence` -> refuter prompt | line 405 |

In fairness, this is **not a new exposure**: the deleted skill also read `git log --oneline` output into context (old `SKILL.md:67,70,100`). What changed is where it lands. The old skill's `allowed-tools` was `Bash Read Grep AskUserQuestion` — one foreground session, under the user's permission prompts. That content now reaches background agents the user never sees.

`opts.agentType` with a read-only tool set is a cheap mitigation and fits the file's own argument at lines 14-17 ("an agent can be talked out of a rule, a regex cannot"). Delimiting the untrusted fields as data would close most of the rest.

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,547 @@
...
+const investigated = []
+const dropped = []
+
+// A refutation counts as refuting unless it says `refuted: false` outright. An omitted
+// or malformed field is not a clearance — this gate has to fail closed, because a wrong
+// "confirmed" costs the user work that exists nowhere else.
+const isRefuted = (x) => x.refuted !== false
+
+results.forEach((r, i) => {
+  const batchBranches = new Map(batches[i].flatMap((u) => u.decide).map((b) => [b.name, b]))
```
hbrodin, line 488: This maps `results[i]` back to `batches[i]`, so the correctness of the whole assembly step rests on `pipeline()` returning index-aligned results — an invariant the script never states.

With a deliberately misaligned stub:

```
ordered   : del=[aa/1..ii/9] (9), unanalyzed=[]
misaligned: del=[bb/2,ee/5,hh/8] (3), unanalyzed=[aa/1,cc/3,dd/4,ff/6,gg/7,ii/9]
```

It fails **safe** — the mismatched branches drop to `unanalyzed` rather than into the delete list — which is the right direction and consistent with the rest of the file's design. Worth a comment recording the dependency, since a future harness change would degrade coverage silently rather than loudly.

plugins/git-cleanup/skills/git-cleanup/SKILL.md
```diff
@@ -1,360 +0,0 @@
----
```
hbrodin, line 1: Removing the last skill leaves the plugin's only entry point unverified by anything in CI. The PR notes the loadability half; the other half makes it sharper:

- `.github/scripts/check_codex_loadability.py:230-254` counts `skills/**/SKILL.md` and compares against loaded skills, so zero skills passes vacuously at `0 == 0`.
- `.github/scripts/validate_plugin_metadata.py` contains **no** references to `commands` at all — the command's frontmatter is validated by nothing.
- `git-cleanup` is now the **only** plugin in the repo with commands and no skills, so this isn't an established pattern being followed.

Between those two, nothing checks that `commands/git-cleanup.md` parses or loads under any CLI, and it's the sole way to invoke the plugin. Agreed the gap is in the checkers rather than this plugin — but I'd rather the loadability check fail on a plugin with no verifiable entry point than pass on zero, and that's a small change worth landing alongside this.

hbrodin (commented):
Review of the analysis path, with the findings I could reproduce by execution. Ran the JS suite (37/37) and the eval self-test (39/39) green on node v24 before starting.

**Two I'd treat as blocking**, both reproduced:
- `git branch -d` accepts branches whose work is *not* in the default branch, so the one delete category that skips refutation has an unsound backstop (`analyze-branches.js:282`).
- `staging`, `production`, `dev`, and `hotfix/*` reach the delete list, with force-delete and an empty `needsReview` on the remote-gone path (`analyze-branches.js:17`).

**Four worth fixing before merge:** the quoting guidance on the agent-facing path, the gate-1 audit rule rejecting the workflow's own `SAFE_TO_DELETE` evidence string, `worktreePath` being optional but load-bearing, and the loadability check now passing vacuously.

**Three more** on subagent tool constraints, unbounded context in the investigator prompt, and an unstated `pipeline()` ordering dependency.

Some things I checked that came back clean, so they don't appear as comments: the refuter prompt not being batch-scoped has no downstream effect (assembly drops out-of-batch verdicts correctly); the `decide`-side prompt cap works as intended at 30 branches per prompt for a 150-branch repo; and a dirty worktree can't actually be removed out from under the user — git refuses without `--force`, and `branch -d/-D` both refuse while a worktree holds the branch. I also measured the fleet cost: a repo with a dozen branches spawns three agents, not eleven, because the deterministic triage decides the rest without spawning anything.

The safety architecture — deletions in the main session, adversarial refutation, triage in JS instead of prose — is the right shape and holds up. The findings below are about places where the verification does less than the surrounding comments claim.

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Fixed in 906aad3 — and you are right that the comment was asserting a property `-d` does not test. I took the third option rather than either of the two you offered: `SAFE_TO_DELETE` now ships a `verifyWith` command, `git merge-base --is-ancestor <tip> <default>`, which the main session runs immediately before the delete and skips on failure. That tests exactly "merged into the default branch", it is deterministic rather than another agent's opinion, and it avoids routing every merged branch through a refutation pass. The evidence string names the tip commit too, so the claim is checkable by a human at gate 1 instead of asserted — which also resolves your gate-1 audit-rule finding.

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Fixed in 906aad3. Your remote-gone reproduction is now a committed test case (case 12), covering both paths and `Staging` for the case-sensitivity hole. The list now covers long-lived integration and environment names — `staging`, `production`, `prod`, `preprod`, `dev`, `qa`, `uat`, `integration`, `next`, `canary`, `stable`, plus `release/*`, `hotfix/*`, `support/*`, `maint*/*` — and matches case-insensitively. Both of your reproductions return empty now.

plugins/git-cleanup/references/merge-evidence.md
frabert: Fixed in 906aad3. Both agent-facing copies now require single quotes and give the `'\'''` escape, and both say why: the agents paste literal names rather than expanding a shell variable, so double quotes still run `$(...)`. The reference file also notes that its own `"$branch"` examples are the variable case and not the one the reader is in. Good catch that `has'quote` is legal — that makes the escape in the command file load-bearing, so it is now stated in all three places rather than one.

plugins/git-cleanup/commands/git-cleanup.md
frabert: Fixed in 906aad3. Rule 1 now says `SAFE_TO_DELETE` satisfies it by naming the tip commit, and the evidence string does name it (`tip abc1234 reported by git branch --merged as an ancestor of main`). So the rule is satisfied on its own terms rather than by an exemption. Rule 2 also stopped naming a four-entry list that no longer matches the regex.

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Fixed in 906aad3. `report()` now builds a `shortRef(branch) -> path` map from the required `worktrees[]` array and prefers it, falling back to the optional field only if the join misses. A schema-conformant survey that omits `worktreePath` now still gets the ordering right, which case 14 pins.

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Fixed in 906aad3. Context is capped at 8 per unit and ranked so tracked siblings come first — a supersession claim needs the branch that is still live, and an unbounded tail of stale local leftovers was the part that multiplied across slices. Thanks for measuring the `decide` side separately; that is what told me the cap belonged only on context.

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Partly fixed in 906aad3, and you have the shape of it right. Untrusted spans — the merge log, the branch/commit lines, and the investigator's `evidence` field flowing into the refuter — are now fenced in `<repo-data>` with an explicit boundary saying the contents are data, cannot lift the read-only constraint or claim a step was already done, and that text reading like an instruction is itself a finding to report.

I did not add `agentType`. The workflow's `agent()` takes no tool list, so restricting tools means shipping a plugin agent and depending on `<plugin>:<agent>` resolving inside a workflow subagent — which I cannot test from here, and an unresolvable `agentType` fails the whole run rather than degrading. I would rather not trade a working analysis for an untested restriction on this PR. Your framing of the real change — same content, but now reaching background agents the user never sees — is the right one, and worth its own change once the resolution behaviour is confirmed.

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Fixed in 906aad3 — recorded as a DEPENDENCY comment at the `results.forEach`, including your finding that misalignment degrades to `unanalyzed` rather than into the delete list. Agreed that is the right direction; the risk was that it degrades silently, so the note says so explicitly.

plugins/git-cleanup/skills/git-cleanup/SKILL.md
frabert: Fixed in 906aad3, in the checkers rather than by restoring the skill — the point of the PR is that workflow-shaped work should be a workflow, so a skill kept purely as a loadability anchor would undo it. Two changes to `validate_plugin_metadata.py`:

- `validate_command_frontmatter` — command files must have parseable frontmatter, a `description:`, and `allowed-tools:` rather than `tools:`. You are right that the validator contained no references to `commands` at all.
- `validate_entry_points` — a plugin exposing no skills, commands, agents, hooks, or `.mcp.json` is now an error, so a plugin that ships nothing runnable cannot pass at `0 == 0`.

Six self-test assertions cover both, including the two negative directions that matter: a valid command is accepted, and commands alone satisfy the entry-point rule. Neither check fires on any existing plugin. This does not make Codex load `commands/` — that is still a real gap — but the plugin's entry point is no longer unvalidated by everything.

frabert:
@hbrodin — thanks, this was a genuinely useful review. All nine are addressed in 906aad3, and I've replied in each inline thread with the specifics; this is the summary in one place since those replies collapse.

**Both blockers were right.**

The `-d` one is the one I'd most wanted caught. I had written a comment claiming `git branch -d` re-derives the merge into the default branch, and it doesn't — merged-into-HEAD *or* merged-into-upstream, as you reproduced. So the one delete category that skips refutation had nothing behind it, and the comment was actively misleading about that. I took a third route rather than either you suggested: each `SAFE_TO_DELETE` candidate now carries a `verifyWith` of `git merge-base --is-ancestor <tip> <default>`, which the main session runs immediately before the delete and skips on failure. That tests the property actually claimed, it's deterministic rather than a second model's opinion, and it costs no extra agents. The evidence string names the tip commit, which also settles your gate-1 audit finding on its own terms rather than by an exemption.

`PROTECTED` now covers long-lived integration and environment names and matches case-insensitively. Your remote-gone reproduction is committed as test case 12, including `Staging` for the case hole; both of your repro paths return empty now.

**The four before-merge items** are all fixed: single-quote guidance mirrored into the two agent-facing copies (with the `'\''` escape, since `has'quote` being legal makes it load-bearing rather than theoretical), the gate-1 rule no longer rejecting the workflow's own evidence, `worktreePath` derived from the required `worktrees[]` array, and the loadability gap closed in the checkers.

On the last of those — I fixed it in `validate_plugin_metadata.py` rather than by restoring a skill, since a skill kept purely as a loadability anchor would undo the point of the PR. Command frontmatter is now validated (the validator previously had no references to `commands` at all, as you noted), and a plugin exposing no entry point is an error, so `0 == 0` can't be the whole story. Six self-test assertions cover both directions, including that commands alone satisfy the rule. It does not make Codex load `commands/` — that gap is real and still open.

**One I only partly did**, flagged so it isn't mistaken for done: the subagent tool constraint. Untrusted spans — merge log, branch and commit lines, and the investigator's `evidence` flowing into the refuter — are now fenced in `<repo-data>` with an explicit boundary saying the contents are data, cannot lift the read-only constraint or claim a step was already done, and that text reading like an instruction is itself a finding. I did not add `agentType`: `agent()` takes no tool list, so restricting tools means shipping a plugin agent and depending on `<plugin>:<agent>` resolving inside a workflow subagent, which I can't verify from here — and an unresolvable `agentType` fails the whole run rather than degrading. Trading a working analysis for an untested restriction felt like the wrong call to make unilaterally. Your framing of what actually changed — same content, but now reaching background agents the user never sees — is the right one, and I'd rather do it properly in its own change.

**Two notes back**, on things from your clean list:

- The refuter not being batch-scoped is safe *because* assembly drops out-of-batch verdicts — and that drop is itself a fix from the previous review round. They're coupled, so if the assembly guard ever loosens, the refuter's scope stops being harmless. Worth knowing before someone simplifies one of them.
- Your fleet measurement (a dozen branches → three agents, not eleven) is the number I should have put in the README instead of the worst-case eleven. Fixed the framing there.

Suite is at 44 assertions, all green, and the two blocker reproductions are committed as cases rather than left as prose. I've left your threads unresolved for you to close as you see fit.

kz-tob:
need the linter regression fixed, and an approval from @hbrodin

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Follow-up in 3e6b2c3: `verifyWith` now names `refs/heads/<branch>` rather than the tip sha the survey reported. The agent joins `branch -vv` and `branch --merged` itself, so a transposed row could carry a sha that *is* an ancestor of the default branch while the branch is not — the precondition would then pass on a branch it never examined, and `-d` accepts it too. A refname cannot desynchronise from the branch it names. The sha stays in the evidence for the human check at gate 1.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/git-cleanup/references/merge-evidence.md
frabert: Follow-up in 3e6b2c3: the single-quote rule now applies to the one shell command this workflow builds *itself*, not just the ones it tells agents to build. `verifyWith` runs both refnames through a new `sq()` helper with the `'\''` escape, and the JS suite pins it — a branch named `evil$(id)` renders as `git merge-base --is-ancestor 'refs/heads/evil$(id)' 'main'`, verified to round-trip through bash unexpanded. Your point that `has'quote` is legal is what made the escape rather than plain quoting the right choice here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

frabert:
Merged `main` (twice — it moved under me) and worked the automated review. hbrodin's nine threads were already answered in `c5358a1`; these are the `github-actions` review's findings, which were not. In `3e6b2c3`.

**Correctness**

- **#4 — `verifyWith` checked an agent-reported sha, not the branch.** Right, and this was the one that mattered. It now reads `git merge-base --is-ancestor 'refs/heads/<branch>' '<default>'`. The survey agent joins `branch -vv` and `branch --merged` into one row itself, so a transposed or stale `lastCommit` could carry a sha that *is* an ancestor of the default branch while the branch is not — the precondition would have passed on a branch it never examined, and `-d` accepts it too. A refname cannot desynchronise from the branch it names. The sha stays in the evidence, where a human reads it at gate 1. This also removes **#12**: no `(unknown)` reaches the command, and a missing tip now reads "tip commit not reported" instead of producing a bash syntax error.
- **#5 — unquoted refnames in the one command the workflow builds itself.** Fixed via an `sq()` helper using the `'\''` escape. Verified round-tripping through bash: `evil$(id)` renders as `'refs/heads/evil$(id)'` and does not expand.
- **#6 — destructive-command regex gaps.** All five spellings you listed now match, plus `git --no-pager branch --delete`. Global options are consumed generically rather than enumerated. The self-test feeds it all eleven destructive forms and three non-delete pushes that must still pass, since broadening a pattern is how you start failing honest runs.
- **#7 — `g_all_branches_mentioned` passed on zero items.** Now ERRORs, same guard as the worktree grader, with an assertion proving it.
- **#8 — fixture inherited the caller's git config while claiming reproducible shas.** `GIT_CONFIG_GLOBAL`/`SYSTEM` to `/dev/null`, hooks and signing disabled per-repo. Worth flagging how bad this one was: `eval-self-tests` is inside `make check`, so anyone with `commit.gpgsign = true` had the whole build blocking on a passphrase. `3fcf672`, `64b5c2a` and `a2f470c` are unchanged.
- **#2 — literal `${CLAUDE_PLUGIN_ROOT}` with nothing to expand it.** The command now resolves it first, and an unreadable `scriptPath` falls through to the inline path rather than aborting.

**Docs that contradicted the code** — #1 (both halves), #9, #10, #11, #13, #15, #16 all fixed. #1 was the worst of them: the README stated the `git branch -d` rationale this PR exists to disprove and never mentioned `verifyWith`, so your failure scenario — a maintainer dropping the precondition as redundant — was live. #16 was the stale side of the Makefile/AGENTS.md disagreement: CI does not scope the validator, only the version-increment check.

**#14** dead `grep -v` removed. **#17** PR body is now accurate against 2.2.0 and the current assertion counts.

**Not addressed — #3, the Codex gap.** You are right that `validate_entry_points` does not close it: it accepts `commands/` regardless of which CLI can load it. Making git-cleanup loadable under Codex means either keeping a skill purely as an anchor — which undoes the point of the PR — or Codex support for `commands/`. That is a maintainer call about plugin shape, and I would rather it be made deliberately than smuggled in here.

Version 2.2.0 in both manifests. Suites: 47 JS assertions, 49 eval self-test assertions, 53 validator assertions.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,636 @@
...
+// JavaScript rather than in an agent prompt: an agent can be talked out of a rule,
+// a regex cannot.
+//
+// These are long-lived integration and environment branches, not feature work. The
+// dangerous case is not that they look merged — it is that their remote gets deleted
+// during a branch-protection change or a repo migration, at which point the [gone]
+// path recommends `git branch -D` on `production`. Names are matched case-insensitively
+// because `Staging` and `staging` are the same branch to everyone except a regex.
+const PROTECTED =
+  /^(main|master|trunk|default|develop|dev|devel|integration|staging|stage|production|prod|preprod|qa|uat|test|testing|sandbox|demo|next|canary|latest|stable|release[/-].*|hotfix[/-].*|support[/-].*|maint(enance)?[/-].*)$/i
```
hbrodin: **The widened list drops branches out of every output list, not just the delete list.**

The filter at line 275 runs before triage and `report()` only reads `settled`/`investigated`, so protected branches land in no output array. Survivable at `main|master|develop|release/*`; not with `test|testing|demo|sandbox|next|latest|default|dev` and `/i`.

Stubbed run with `test` (tracking a remote, **7 unpushed commits**), `demo`, `next`, `Latest`, `dev`:

```
logs: 6 local branches; 6 protected or checked out, 0 in scope
deleteCandidates: 0  needsReview: 0  keep: []  unanalyzed: []
```

Unpushed work, gone from the report with no trace — not into `keep`, not into `unanalyzed`, which is what Safety Rule 7 relies on to stop a partial analysis reading as complete. Every other failure direction here costs a second look at a branch list; this one shows a quietly incomplete picture.

1. Surface them: a `protected: [...]` array, or `keep` with `category: 'PROTECTED'`. Never deletable and never mentioned are different guarantees; `commands/git-cleanup.md:82` currently asks for the second.
2. Trim the regex. The integration/environment names earn their place — the `[gone]`-on-`production` scenario is exactly right. `test`, `testing`, `demo`, `sandbox`, `latest`, `default` don't, and `/i` takes `Test` and `Demo` too.

Part 2 is a judgement call and goes away if you read those as environment branches in the target repos. Part 1 stands regardless — the same silence hits `staging` with unpushed work on it.

plugins/git-cleanup/workflows/analyze-branches.js
```diff
@@ -0,0 +1,636 @@
...
+// failed to read anything, which must not be reported as "nothing to clean up".
+if (!survey.branches || survey.branches.length === 0) {
+  throw new Error('survey returned zero local branches — the inventory failed rather than finding a clean repo')
+}
+
+// The default branch is protected whatever it is called: `git branch --merged trunk`
+// lists `trunk` itself, so a repo not using one of the conventional names would
+// otherwise see its own trunk recommended for deletion.
+const isProtected = (name) =>
+  PROTECTED.test(name) || name === survey.defaultBranch || name === survey.currentBranch
```
hbrodin: **This comparison misses, because line 233 doesn't emit a branch name.**

`git symbolic-ref refs/remotes/origin/HEAD` with no `--short` prints `refs/remotes/origin/mainline`. `SURVEY_SCHEMA.defaultBranch` (line 104) has no `description` steering the agent to the short form either. For a default branch outside `PROTECTED`, the repo's own trunk lands on the delete list:

```json
{ "branch": "mainline", "category": "SAFE_TO_DELETE", "command": "git branch -d",
  "verifyWith": "git merge-base --is-ancestor 'refs/heads/mainline' 'origin/mainline'" }
```

`verifyWith` returns 0, so this round's precondition doesn't catch it. Verified on a repo actually built with `mainline` as default — `git branch --merged refs/remotes/origin/mainline` still lists `mainline`, so the survey stays self-consistent throughout.

Narrower than it looks: `--merged` only lists a branch whose tip is an ancestor, so if local `mainline` has commits origin lacks it never becomes a candidate *and* `verifyWith` returns 1. The delete only fires when local is level with or behind origin, recoverable via `git switch mainline`, and needs the user checked out elsewhere. Recoverable annoyance, not lost work — but the correct implementation is already in this PR: the command file's inline fallback does `symbolic-ref --short` then `${default_branch#origin/}`, so the two analysis paths disagree.

```js
// shortRef only strips refs/heads/; the default branch arrives as a remote ref
const localName = (ref) => String(ref || '').replace(/^refs\/(heads|remotes)\//, '').replace(/^origin\//, '')
```

Apply to `defaultBranch` and `currentBranch`; use in `isProtected`, `verifyWith` (line 343) and the prompts; `--short` at line 233; `description` on both schema properties. `currentBranch` has the same exposure, failing safe today only because `-d` refuses the checked-out branch.

Unprotected real-world default names: `mainline`, `development`, bare `release` (pattern requires `release[/-]`), `unstable`, `primary`, `current`, `live`, `gh-pages`, `edge`, `upstream`.

hbrodin (commented):
Re-reviewed only what changed since my last pass. Every one of the nine findings from that round is addressed, and I re-verified each by execution rather than reading the replies — `verifyWith` naming `refs/heads/<branch>`, the environment-branch filter, the `worktreeOf` join, the capped-and-logged context, the recorded `pipeline()` alignment dependency. Locally green: 47/47 JS assertions, 49/49 eval self-test, 82/82 validator self-test, validator clean across all 41 plugins.

Two findings below, both reproduced by running the script with stubbed agents. Both are in the deterministic core rather than in agent behaviour, so both are cheap to pin with a test.

Things I checked that came back clean, so they aren't comments: `sq()` correctly escapes `$(...)`, backticks and embedded single quotes in `verifyWith`; detached-HEAD worktrees don't collide in the new `worktreeOf` map; the `GIT_CONFIG_GLOBAL=/dev/null` fixture hardening doesn't break repo creation; and the new `run-evals.sh --self-test` exec path works on a machine without `claude` installed. I also confirmed the `<repo-data>` fence is escapable by a branch name (`x</repo-data>_...` is a legal refname) — the ceiling is one ambiguous-bucket branch behind both gates, and everything I threw at it aimed at protected or already-settled branches was dropped by the batch-scoping guard, so I'm raising it as a one-line hardening suggestion rather than a finding: strip the delimiter inside `fence()`.

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Fixed in 0e5d0bf. Took both parts. Protected branches now travel to `report()` and come back under `keep` with `category: 'PROTECTED'`, evidence naming why they were excluded plus the unpushed count when there is one, and an unpushed count on a protected branch is logged separately. And the trim: `test`, `testing`, `demo`, `sandbox`, `latest`, `default` are out — you are right that they are the throwaway names this tool exists to clean up, and over-protection is not free just because it errs safe. Pinned by three new assertions, including `staging` with 7 unpushed commits surviving into `keep`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/git-cleanup/workflows/analyze-branches.js
frabert: Fixed in 0e5d0bf, using your `localName()` more or less verbatim — applied to both `defaultBranch` and `currentBranch`, plus `--short` in the survey prompt and a `description` on both schema properties. The detail that made this worth pinning rather than patching is the one you noted: the survey stays self-consistent throughout, so nothing downstream contradicts it and `verifyWith` returns 0 too. Regression case runs all three reported spellings (`refs/remotes/origin/mainline`, `origin/mainline`, `mainline`) and a fully qualified `currentBranch`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

kz-tob (approved):

PR #219, opened 2026-07-30
Add first-pass eval suite for yara-authoring

kz-tob:
Eight cases measuring rule tightness for the yara-rule-authoring skill, scored as uplift over a no-plugin baseline arm.

Six cases where the skill should fire: prose intel writeup, raw string dump, IOC bulletin, generic-indicators-only, over-broad category ask, and binary triage from a base64-embedded synthetic PE. Two cases where it should not: a Semgrep request and a log-parsing request.

Graders favour verifiable checks (regex over the produced .yar file, file_exists) with LLM rubrics reserved for properties regex cannot express, notably whether any generic string can trigger a match without a family-unique indicator also matching. Every case carries at least one outcome grader; the tool_used Skill checks are display-only and excluded from scoring.

Malware families and indicators are invented. Hosts use .invalid, the binary fixture is a 560-byte synthetic PE that is not executable code, and no real sample is included. Invented families also keep the baseline arm from scoring on recall of published rules.

Run with:
  claude plugin eval . --ablation with-without --judge-model opus \
    --allow-tools Bash Write


KernelClint:
## Partner review — ran it cold

Fresh clone, 11 eval runs ($2.60), plus ~24 mutations against the Python tests. Short version: **case 02 is a good test. Case 04 isn't. The merge gate isn't proven yet.**

I also compared my eval run against the HTML report Kara shared. They look different but they don't disagree — see below.

### Case 02 works. Case 04 measures nothing.

Same commit, both runs 12 minutes apart. Pooling them:

| | report | my runs | pooled |
|---|---|---|---|
| Case 02 with plugin | 100, 100, 100 | 100, 82, 82 | 94% |
| Case 02 baseline | 45.5, 63.6, 45.5 | 63.6 | 55% |
| **Case 02 Δ** | **+48.5** | +36 | **+39** |
| Case 04 with plugin | 50, 100, 100 | 100, 100 | 90% |
| Case 04 baseline | 50, 50, 100 | 100, 100 | 80% |
| **Case 04 Δ** | **+16.7** | **0.00** | **+10** |

Case 02's uplift is real. The baseline arm failed `junk-cannot-fire-alone` and `explains-rejections` in 4 of 4 runs on both machines. +48.5 is a bit lucky; ~+39 is the honest number.

Case 04's Δ is noise. Look at the report's own arms: with-plugin scored 50/100/100, baseline scored 50/50/100. Same two values, different order. The +16.7 comes down to one baseline run landing at 50%. All four of my runs scored 100/100. Suggest replacing this case, and piloting any replacement against the baseline arm before keeping it.

### The merge gate needs one more experiment

The rule is "a test case that fails when the skill is broken." So I deleted 1,540 chars of SKILL.md: the string-rejection table, the goodware-validation principle, the matching Rationalizations row, and I flipped the `all of` / `any of` table to say "always use `any of them`". Then ran case 02 gutted vs intact, same config:

| skill | run 1 | run 2 |
|---|---|---|
| intact | 0.83 | 0.83 |
| gutted | 0.83 | 0.83 |

Same score. But that's not quite "the suite can't detect breakage." Two things are going on:

- `explains-rejections` is too noisy to gate on. It passes 4 of 6 with-plugin runs, passes 1 of 2 gutted runs, and it failed *both* my intact runs while passing all three in the report. Same skill, different answers.
- `junk-cannot-fire-alone` is the grader doing the real work: 6/6 pass with the plugin, 0/4 at baseline. That's the cleanest split in either run. It passed gutted too, but I think that's my fault — I flipped the `any of` table but left Core Principle 2, the FP-prone strings table, and Grouping by Confidence alone, so the skill still had enough to write a tight condition.

So the test that actually matters hasn't been run yet: strip everything backing `junk-cannot-fire-alone` at once and see if it moves. Doing that now and will post the result.

### Two grader bugs

**`filesize-prefilter` passes wrong answers and fails right ones.** Every case-02 run in the report passed it with a bound nowhere near the ~340KB sample: `2MB`, `1MB`, `5MB`. The README says this grader exists so an overfit rule doesn't score the same as one reasoning from the stated size — but all three perfect runs used 3–15x the sample size and got full marks, because `\d+\s*MB` accepts any number. Meanwhile it *rejects* `filesize < 400_000`, the underscore form SKILL.md:268 recommends, and `filesize < 0x80000`. Needs an upper bound plus the `_` and hex forms.

**`pe-magic-prefilter` rejects `uint16be(0) == 0x4D5A`** — the byte-order spelling SKILL.md:63 tells you to use. One more alternation fixes it.

### The eval depends on what's installed

The report's runs say `yr check` and `yr fmt --check` pass, and show `yr scan -s`. My runs said "`yr check` — I hand-verified the syntax instead," because `yr` isn't on a clean machine. One agent could compile and scan its rule; mine couldn't. That probably explains part of the stronger case-02 scores, and it means scores aren't comparable between machines — which is the whole reason for keeping the case numbers stable. Either list `yr` as a prerequisite or install it in a `scaffold_script`.

Also: the run command in `evals/README.md` fails with `plugin eval is currently in early access` unless you set `CLAUDE_CODE_WALNUT_SPIRE=1`. Worth adding.

### Python tests: linter is well covered, atom analyzer isn't

34/34 pass. All 14 mutations I made to the lint checks were caught — E002, E003, E008, W002, W004, W005, W006, W008, the description limits, hex-run splitting. The style-guide table sync test does real work.

The atom analyzer is the weak half. 10 mutations survived:

- `score_atom` has no test holding it in place. I zeroed the null-byte, ASCII, common-sequence and repeated-pattern penalties one at a time — all 34 tests still passed. Both thresholds (30 and 60) can be set to 0 and nothing fails.
- `strip_comments` block-comment handling and the regex scanner's character-class tracking both have no test that fails when you switch them off. These are the two scanners the module docstring specifically explains why it hand-wrote.
- `select_exit_code` never tests the error path. The existing test covers clean, warnings-only, and unreadable files — but not "an error exits 1", which is the linter's main job.

### One real bug

`atom_analyzer.py` reports success on a file that doesn't compile:

```
$ uv run atom_analyzer.py notes.yar     # contains prose, not a rule
✓ All strings in notes.yar have good atom quality
$ echo $?
0
```

The compile error only goes to stderr. Then the rule loop never runs, so `has_issues` stays false. `yara_lint.py` gets this right and exits 1. `collect_rule_files` guards directories but passes single files through unchecked, so `README.md:63` isn't accurate either — the non-zero-on-nothing-inspected claim only holds for directories.

Related and pre-existing: escape sequences aren't decoded, so `$s = "\x41\x42"` counts as 8 bytes instead of 2. E002 stays quiet, and the analyzer suggests atom `5C783431` — bytes that exist in no file.

### Two items from the bot review I'd skip

- **`uint64` W009 false positive** — not a bug. YARA-X has no `uint64` or `int64`; the compiler rejects both. `CHEAP_CHECK_RE` is right to leave them out.
- **Breaking code renumbering** — not a bug. I diffed old against new: E002, E003, W003, W004 and W006 mean exactly the same thing. Only E007 was retired (deliberately, superseded by E000) and W009 added. The commit message line "Linter codes are renumbered" is the inaccurate part, not the code.

Also checked: this merges cleanly against current `main`. The `workflow-skill-design` removal (#215) and the `skill-improver` 1.0.5 bump both survive, despite how the diff looks.

### Suggested order

1. Replace case 04.
2. Run the targeted ablation on `junk-cannot-fire-alone`. In progress.
3. Fix `filesize-prefilter` and `pe-magic-prefilter`. Report run 1 is your regression test.
4. Raise case 02 to `runs: 5`, or swap `explains-rejections` for a mechanical check — does the reply name at least 2 of the 9 junk strings.
5. List or install `yr`; document `CLAUDE_CODE_WALNUT_SPIRE=1`.
6. Fix the `atom_analyzer.py` exit code; add the missing `select_exit_code` test.
7. Small stuff: SKILL.md still has `When to Use` / `When NOT to Use`, which `main` dropped in #216 — 18 free lines. The PR description still describes 8 cases and a PE fixture that aren't in the tree. `crx-module.md:67` compares `permhash()` against 64 zeros, which compiles and can never match.

`license/cla` is still pending, separate from all of this.

Holding approval on 2 and 3 only. The rest I'd take as follow-ups. The skill is clearly better than it was, and case 02 is a real test rather than a box-tick.

KernelClint:
## Ablation result: the suite does not detect a broken skill

Ran the experiment I promised above. Answer is clear, and it isn't noise this time.

I cut 2,988 chars from SKILL.md — everything that teaches "generic strings must not fire alone":

- Core Principle 2 ("Target specific families, not categories") and 3 ("Test against goodware before deployment")
- the `**Value ranking:**` line (mutex gold / C2 silver / error messages bronze)
- the whole string-rejection table, replaced with "Use whatever the extractor gave you; more strings means better coverage"
- the `all of` vs `any of` table, inverted to a single row: "Any strings at all → `any of them`, keeps the rule flexible"
- the entire `### Grouping by Confidence` section
- the "Lesson from production" FP paragraph
- 4 Rationalizations rows, including "`any of them` is fine for these common strings" and "This generic string is unique enough"
- the "Multiple generic strings matching together?" tightening step

Both arms run from isolated copies of the plugin, `--runs 3 --ablation none --judge-model opus`:

| skill | run 1 | run 2 | run 3 | mean |
|---|---|---|---|---|
| intact (clean checkout) | 0.83 | 1.00 | 1.00 | **0.94** |
| gutted (−2,988 chars) | 0.92 | 1.00 | 1.00 | **0.97** |

The gutted skill scored slightly **higher**. A fourth gutted run scored 1.00.

`junk-cannot-fire-alone` — the grader with the cleanest baseline separation — passed **7 of 7** plugin-present runs, every one a unanimous 3–0 judge vote, whether or not the guidance behind it existed.

### What the grader is actually measuring

It separates *a skill loaded* from *no skill loaded*. It does not separate *good skill* from *gutted skill*.

- no plugin at all: fails 4/4 (3 runs in the HTML report, 1 in mine)
- plugin loaded, guidance intact: passes 3/3
- plugin loaded, guidance deleted: passes 4/4

So the +39 uplift is real, but it comes from loading *something* named `yara-rule-authoring` whose description mentions false-positive reduction. The body text isn't carrying it. Case 02's prompt hands over 3 obviously-unique indicators (a mutex, a PDB path, a C2 path) and 9 obviously-generic ones — once the model is in "be careful about YARA" mode, the right split is inferable from the prompt alone.

That's why Dan's gate can't be met by tuning `runs:` or rewording these rubrics. The case tests judgment the base model already has.

### What would actually fail when the skill breaks

Test facts the skill teaches that a base model gets wrong. These are regex-checkable, need no LLM judge, and are cheap:

| Fact | Wrong answer a base model gives | Where the skill teaches it |
|---|---|---|
| Mach-O universal binary magic | `uint32(0) == 0xCAFEBABE` — compiles, never matches | SKILL.md:63, Common Mistakes |
| ZIP/OOXML magic | `uint32(0) == 0x504B0304` — same dead branch | Platform table, SKILL.md:58 |
| DEX file-type gate | `dex.header.magic == "dex\n"` — will not compile | `references/dex-module.md` |
| Negative string index | `@a[-1]` — unsupported in YARA-X | Migration table |
| base64 modifier | accepts strings under 3 chars | Migration table, E006 |

A "migrate this legacy rule to YARA-X" case with several of these planted would fail loudly the moment the migration table or the endianness callout is removed — which is exactly the property the merge rule asks for. Happy to draft it if useful; I'm building the grader fixes now either way.

### Method note

My first pass at this was wrong in two ways and I want it on the record. I ran an earlier 2-run comparison that came out 0.83 vs 0.83 and read it as "the suite can't tell" — but that ablation only removed 1,540 chars and left Core Principle 2, the FP-prone table, and Grouping by Confidence in place, so it wasn't a fair test of `junk-cannot-fire-alone`. Separately, one intact 3-run eval read from the live `plugins/yara-authoring` directory while I was editing scripts and running mutation sweeps in it, so that 0.97 was contaminated. Both arms above were re-run from isolated copies, the intact one straight from `git archive`. The numbers in this comment are the ones to use.

KernelClint:
Put the fixes on a branch for you to pull from: **#229**, based against `evals/yara-authoring-first-pass` so the diff is only my changes. Cherry-pick, merge, or ignore.

Two things landed since my last comment:

**`explains-rejections` is fixed, and the cause wasn't judge temperature.** It conflicted with `junk-cannot-fire-alone`: that grader accepts generic strings as corroboration gated behind a unique indicator, while `explains-rejections` demanded two strings be dropped outright. A run that gated everything instead of dropping it satisfied one and failed the other — that's the PASS/FAIL/FAIL on identical runs. It now takes either disposition as long as the reply says which and why. Measured over 3 runs per arm: 3/3 pass with the plugin, 3/3 fail without, every vote unanimous, baseline pinned at 0.45 on all three runs. **Case 02 goes Δ +0.39 → +0.52.** Runs raised to 5.

I tried the mechanical swap I suggested and it does not work — rejection-language plus two junk strings passes all three of your recorded with-plugin replies but also two of three baseline replies, because the baseline names the junk, calls it generic, then uses it anyway in a fallback branch. Checking the rule file fails too, since gated corroboration is allowed. Both dead ends are written into the README so nobody retries them.

**`06-review-with-linter` is the gate.** `reports-linter-codes` passes 3/3 intact, 0/3 with the Scripts section removed, 0/2 with no plugin. Regex, so no judge variance. Building it also turned up a real gap: the skill directed linter use when authoring but not when reviewing, so it only ran the linter in half of review runs. Adding a review path took that grader from 1/2 to 3/3.

On your DM question — I don't think this is a delete. Case 02 at Δ +0.52 with a stable baseline is a genuine result, and case 06 gives you something that actually fails when the plugin breaks. What I'd drop is the assumption that 02 and 05 gate anything; they measure that a skill loaded, not what it says.

KernelClint:
## Short version

Everything below is in **#229** — pull whatever you want from it.

### What I found

I ran your suite cold, then deleted big chunks of SKILL.md and ran it again. It scored the same:

| Case | intact | with the guidance deleted |
|---|---|---|
| 02 | 0.94 | 0.97 |
| 04 | Δ 0.00 in both arms | — |

So the suite was measuring whether *a skill loaded*, not whether the skill is any good. Sonnet already knows `any of` across generic strings is loose — it just needs to be in a YARA frame, and the `description` alone does that.

### What fixes it

A case that tests something the base model can't fake: your linter's issue codes.

`06-review-with-linter` hands over a bad rule and checks the reply reports `E002`/`W009`-style codes. It passes 3/3 with the skill, **0/3 with the Scripts section removed**, 0/2 with no plugin. It's a regex, so no judge noise. That's a real gate.

Building it found a genuine gap too: the skill told you to run the linter when *writing* a rule but not when *reviewing* one, so it only ran it half the time. I added a review path — that grader went 1/2 → 3/3.

### Other changes

- **`explains-rejections` stopped flipping.** It contradicted `junk-cannot-fire-alone`: one grader allows generic strings as gated corroboration, the other demanded they be dropped entirely. Fixed the contradiction → 3/3 pass with plugin, 3/3 fail without. **Case 02 Δ +0.39 → +0.52.**
- **Two graders rejected correct answers.** `filesize` failed `400_000` (the form SKILL.md:268 recommends) and `0x80000` while accepting `900MB`. `pe-magic` failed `uint16be(0) == 0x4D5A`.
- **Deleted case 04** — Δ 0.00 over four runs.
- **`atom_analyzer.py` said "good atom quality" and exited 0 on files that don't compile.** Fixed.
- **12 new tests.** `score_atom`'s penalties could all be zeroed with all 34 tests still passing.

### Two bot-review items to skip

`uint64` isn't real in YARA-X (compiler rejects it), and the linter codes weren't renumbered — only E007 retired, W009 added.

### On your DM

I wouldn't delete this plugin. Case 02 at Δ +0.52 with a stable baseline is a real result, and case 06 gives you something that actually breaks when the plugin breaks.

KernelClint (approved):
Approving. Partner validation is done — ran the suite cold from a fresh clone, mutation-tested the Python suite, and measured the skill against deliberately gutted copies of itself.

The findings and fixes went into #229: 06-review-with-linter as a regression gate that actually fails when the skill breaks (3/3 with the skill, 0/3 with the Scripts section removed), two graders that were rejecting correct answers, explains-rejections stabilised after finding it contradicted junk-cannot-fire-alone, case 04 removed for measuring Δ 0.00, the atom_analyzer.py exit-0-on-broken-files bug, and 17 new tests closing every mutation that previously survived.

#229 got a full bot review, @kz-tob approved it at 40d67fb, and it's now squash-merged into this branch as 8ee6493, so the diff here includes all of it.

One disclosure: commit f197338 landed after that approval. It fixes a regression I introduced in my own earlier commit — a per-file zero-rule guard that made atom_analyzer.py exit 1 on healthy ruleset directories containing an imports-only file — and corrects a claim I'd made in the eval README that the linter codes "cannot be guessed" (they're tabulated in references/style-guide.md). It's squashed into 8ee6493 and no one but me has read it. It only touches problems in my own preceding commits, CI is green on it, and I verified the behaviour both directions, but it hasn't had a second pair of eyes.

Everything verifiable passes: 51 tests, ruff clean, make validate clean (345 references resolved), SKILL.md 436 lines. All seven CI checks SUCCESS.

Good to go from my side.

PR #220, opened 2026-07-30
Fix constant-time analyzer backends and trim the skill

jvdprng:
Five commits, each independently testable. Reviewing commit by commit is much
easier than reading the combined diff.

- `467a8a5` — 13 per-language triage fixtures. Verdicts are deliberately absent
  from the source; the fixtures double as evaluation input.
- `308f050` — ty configuration and the typing weaknesses it reported. No
  behaviour change.
- `d21ed03` — the backend fixes. This is the commit that needs scrutiny.
- `a2e0ad3` — `expectations.json` (53 cases across 13 languages) and the test
  that asserts every true/false-positive pair still holds.
- `455aeba` — SKILL.md trim, README corrections, version bump to 0.2.0.

## What was wrong

Six backends skipped unparseable toolchain output and reported PASSED on
vulnerable code. Each was covered by tests that replay hand-written listings,
which is why it went unnoticed:

- **Python** — the `BINARY_OP` oparg map read 12 as `//` when 12 is `^`, so floor
  division on a secret was missed and every constant-time XOR was reported as a
  division. The instruction regex also required a byte offset 3.13 no longer
  prints. On 3.13, `key_coef // (2 * gamma2)` came back PASSED.
- **PHP** — the opcache parser delegated to the VLD parser for a format that
  shares nothing structural with it, and `analyze()` called the VLD parser
  unconditionally, so the function was dead either way. VLD is an unbundled PECL
  build, so this is the common path: zero findings on a file with `intdiv`,
  `mt_rand` and `base64_encode`.
- **JavaScript/TypeScript** — the V8 line format meant every file parsed as zero
  instructions. Also compiles eagerly now, filters node's internals, and converts
  V8's byte offsets to line numbers.
- **Go** — findings were dominated by the linked-in runtime (23 of 24), and Plan 9
  syntax writes a 32-bit divide as `SDIVW`, which was absent from the table, so
  `int32` division read as clean.
- **Rust** — compiled as a bin crate, so any library file failed E0601.
- **Swift** — target triples were hardcoded to Apple platforms; no Swift file
  could be analyzed on Linux.

All six source-level scanners also counted comment text as code, and the text
report labelled scripting-language errors `[WARN]` while the summary counted them
as errors.

## Two decisions worth a second opinion

- PHP and Go now raise instead of returning PASSED when nothing parsed. An empty
  parse is "we understood nothing", not a clean bill of health — but it is a
  behaviour change for any caller treating PASSED as success.
- The V8 filter matches declared function names, since V8's dump has no file
  attribution. It over-captures on purpose: an extra name admits a same-named
  node internal, whereas a missed name drops the user's own function.

The 53 verdicts in `expectations.json` are human-assigned and the one thing no
test validates. Two record measured behaviour over intent, and say so inline:
Rust claims no branch true positive, and Go's branch false positive is its
stack-growth check.

## Verification

All 11 backends were exercised against real toolchains (Go, Rust, Swift, Kotlin,
Java, PHP, .NET + ilspycmd, Node, TypeScript, Python, Ruby, plus x86_64 and
riscv64 cross-compilers). 89 tests, 101 subtests. Every behavioural test was
checked by reverting its fix and confirming it fails. `make check` targets, all 15
pre-commit hooks, both loadability checks under the real Claude Code and Codex
CLIs, and `ty check` are clean; the suite and ty pass at each of the five commits.

The SKILL.md changes come from evaluating the skill at three effort levels rather
than from review: all three recommended fixing a secret-dependent divide by
handing the compiler a constant divisor, and measurement showed that fix still
emits a real divide on gcc riscv64 at every optimization level, gcc arm64/x86_64
at Os and Oz, and clang arm64 at O0 and Oz.


tob-scott-a (dismissed):

tob-scott-a (approved):

kz-tob (approved):
LGTM

PR #223, opened 2026-07-30
Trim dwarf-expert: single-file skill, effort level, contract test

ekilmer:
Maker-week verdict for `plugins/dwarf-expert`: **Trim it**.

## Why

- The technical content held up: every llvm-dwarfdump flag the skill documents exists in the real tool, so deletion wasn't warranted. No neighboring plugin covers DWARF, and the skill holds no fixed plan or subagent fan-out, so merge and workflow conversion don't fit either.
- What needed trimming: `reference/coding.md` was mostly generic behavior rules ("Follow Existing Styles", "Group Changes", "Only Suggest Changes") of the kind the Opus 5 prompting guide says to remove; the "When to Use / When NOT to Use" sections duplicated the description (requirement dropped in #216); `reference/readelf.md` was a 9-line file behind a reference hop; an ASCII decision tree restated the section list; five typos.
- Found along the way: the validator's reference extractor matches `references/` (plural) only, so this skill's `reference/` (singular) links were never checked. Consolidation moots it here; the extractor gap may deserve its own follow-up.

## What changed

- **SKILL.md consolidated from 4 files to 1** (115 lines): scope boundaries, authoritative sources, dwarfdump flag reference with implementation disambiguation, search escalation (`--find` → `--name` → grep pipeline → pyelftools), verification workflows, a macOS dSYM note, readelf inlined, the library table, and DWARF-specific pitfalls (optional attributes, `DW_AT_abstract_origin`/`DW_AT_specification` chains, type chains). `reference/` deleted.
- **DWARF version scoping removed everywhere.** The original's "v3-v5, v1/v2 out of scope" was an invented boundary: the guidance is version-agnostic, and evaluation runs handled v2 binaries without it. The version judgment that earned its place lives in the Verifying section — a current-generation compiler emitting old DWARF means the build passed `-gdwarf-N` (gcc embeds the flag in `DW_AT_producer`; clang's producer carries no flags), and v2 output differs only in surface forms (`DW_OP_plus_uconst` member offsets, `DW_AT_MIPS_linkage_name`). Descriptions in `plugin.json`, `marketplace.json`, and both READMEs match.
- **Frontmatter**: adds `effort: medium`; description rewritten around concrete triggers (`.debug_*` sections, DIE trees, `DW_TAG_`/`DW_AT_`, tool names); `allowed-tools` gains Write/Edit/WebFetch — the skill's own escalation path writes Python scripts and its sources section fetches dwarfstd.org.
- **Fixed a factual error carried over from the original**: the DIE-search example told the reader to print matched DIEs with `--lookup=<offset>`, but `--lookup` takes a program address; printing the DIE at a section offset is `--debug-info=<offset>`. Verified live: `--debug-info=0x53` prints the DIE, `--lookup=0x53` prints nothing.
- **New contract test** `tests/test_skill_contract.py` (stdlib + pytest, same shape as rust-review's `test_prompt_regexes.py`: extract claims from SKILL.md, verify against an oracle): every dwarfdump flag the skill documents must exist in a live `llvm-dwarfdump --help`. The resolver picks the newest LLVM on PATH, rejects libdwarf's same-named tool via `--version`, and enforces an LLVM 19 floor — `--error-display` and `--verify-json` are absent from `release/18.x` llvm-dwarfdump.cpp and present in `release/19.x`. Zero extracted flags or a missing tool fails rather than skips. A second test enforces the frontmatter contract (name, description, effort level, space-delimited `allowed-tools`).
  Mutation proof: documenting `--bogus-flag` turns the suite red; `effort: turbo` turns it red; reverting restores green.
- **CI**: the python-tests job installs `llvm-19` in its existing apt step (ubuntu-24.04's default llvm is 18), falling back to the default `llvm` on future runner images that no longer package it. The test caught this live: the first CI run failed against LLVM 18 with a message correctly attributing the failure to toolchain age.
- **Version 1.0.1 → 1.1.0** in plugin.json and marketplace.json; plugin README rewritten; root README row updated.

## Evaluation behind the content choices

Fourteen cold solver runs (Opus 5 / Sonnet / Haiku, each with and without the skill, on macOS/clang and Linux/gcc Debug builds of pe-parse) answered eight DWARF questions graded against a source-derived answer key:

- macOS/clang, where dwarfdump is the only practical tool: no correctness difference at any tier.
- Linux/gcc, where readelf/objdump/gdb compete: Haiku scored 6/8 with the skill at 27 tool calls vs 5/8 without at 83 tool calls (the control drifted into readelf+grep pipelines and misread DIEs); Sonnet tied at 8/8 with 45% fewer tool calls when guided by the skill.
- Three content decisions came directly from observed failures: dropping the version scoping (the first real binary tested was v2 — pe-parse pins `-gdwarf-2`), the producer/version inference line (3 of 6 macOS runs misattributed v2 to a toolchain default), and the macOS dSYM note.

## Verification

- `make check` green: validator self-test, ruff, shellcheck/shfmt, bats, python tests including the new suite, validator (no errors, 359 references resolved)
- `prek run -a` green, including actionlint and zizmor on the workflow edit
- Both loadability checks pass by hand (Claude: 43 plugins; Codex: 43 plugins, 82 skills)
- All CI checks green on this branch, including the python-tests job running the new suite against llvm-19

## Merge gate

Per the hackathon rules this needs a partner cold-run; the break-detecting test is included and mutation-proven above. Draft until then.


ekilmer:
Response to the automated review — two findings fixed in 6d36412, one deferred deliberately, one resolved by citation, two declined with rationale.

**P2 (fence-blind flag extraction) — fixed.** `body_sections()` now tracks fence state and ignores fences only for heading detection; fenced content stays in the section text, so flags documented inside examples are extracted and verified. A new unit test (`test_body_sections_are_fence_aware`) asserts both properties, and the mutation proof now covers this path: appending a fenced ` ```bash llvm-dwarfdump --fenced-bogus-flag file``` ` block turns the suite red.

**P3 (Apple LLVM version floor) — fixed, per your suggested direction.** The parsed-major floor is gone; the resolver now gates on capability — `--error-display` and `--verify-json` must appear in the candidate's `--help` — since Apple's LLVM numbering does not track upstream and a numeric major is not comparable across toolchains. The failure message lists every rejected tool with its version banner. README wording updated to match.

**P3 (AGENTS.md prerequisite note) — deferred, deliberately.** Maintainer call for this PR: the suite's failure message self-documents the prerequisite (what to install on macOS and Debian/Ubuntu), and constant-time-analysis's heavier toolchain requirement is likewise not in AGENTS.md today. Happy to add a sentence in a follow-up if the precedent should change.

**P3 (`effort:` legitimacy) — confirmed supported.** The official Claude Code skills documentation's frontmatter reference lists `effort` as a SKILL.md field with values `low`/`medium`/`high`/`xhigh`/`max` (alongside `model`, `context`, etc.). Setting it is a deliverable of this trim, so the test pinning its presence is intentional, not incidental.

**P4 (single-dash flags escape) — declined.** The skill's Verifying section legitimately mentions `-gdwarf-2`/`-gdwarf-N` — compiler flags, not dwarfdump flags. A single-dash-inclusive regex would extract them and falsely check them against dwarfdump's `--help`. Double-dash-only is the deliberate tradeoff, documented in the comment above `FLAG_RE`.

**P4 (floor is not a coverage check) — partially taken.** Floor raised from 10 to 15 (21 flags currently documented), which fires on the halved-reference scenario. Pinning the exact set is declined: it turns every legitimate flag addition or removal into a test edit, trading rot detection for change-detector churn.

xintenseapple (approved):

kz-tob (approved):

PR #225, opened 2026-07-31
static-analysis: make the codeql skill's guards real and add tests

kumarak:
The skill's verification steps could not fail. Each piped its command to a formatter and
used the pipeline's exit status, which without `pipefail` belongs to the formatter. Nine
sites. Nothing caught it because `make shell` lints `*.sh` files and the skill shipped none.

## Bugs fixed, each verified against CodeQL 2.25.6

- **arm64e detection could never fire.** `EXIT_CODE=$?` after `cmd | tee` reads tee's status
  (always 0), never 137. Essential Principle #5 and three Rationalizations depend on it.
- **Quality gate failed healthy databases.** Diagnostics counted by lines starting with `{`;
  the files are pretty-printed, so nested braces counted as records. A clean database reported
  500% error ratio while CodeQL recorded `extractor-failures: 0`.
- **Go classified as no-build.** It is compiled and rejects `--build-mode=none` outright. The
  interpreted/compiled framing was wrong; re-sorted by build mode.
- **"Run all" resolves 219 of the pack's 515 alert queries**, not "all queries from all
  installed packs" — and 11 of those 219 raise no alerts at all. It leaves 307 unrun,
  including 13 `Security/CWE/` queries. Measured against `cpp-queries` 1.8.0.
  `important-only` is also not a subset of it — different selection mechanisms.
- **Production `analyze` command word-split on spaces.** `$DB_NAME` unquoted; flag bundles
  converted to arrays so they can vanish when empty.
- **Multi-database selection offered one option.** `for db in $FOUND_DBS` loops once.
- **Two bash blocks could not parse** — an `else` branch of only comments.

## Also

Five scripts now carry logic that used to live only in markdown, where `make shell` could
not lint it and no test could run it: `check_db_quality.py` and `verify_query_suite.py`
enforce what the prose claimed, `build_log.sh` holds the logging helpers, `generate_suite.sh`
replaces suite-generation bash that was copy-pasted into two reference docs, and
`find_databases.sh` replaces a discovery loop whose bash array a *different* markdown block
tried to read — arrays do not survive between Bash calls, so it reported "no database found"
for projects that had one.

Three approval prompts collapsed to one. `TodoRead` and `TodoWrite` are dropped from
`allowed-tools`; `TaskGet` stays, since it exists and the skill asks the model to use the
task tools.

The skill's markdown went from 2867 lines to 2744, and `SKILL.md` from 269 to 257.

## Tests

Nine test files in `scripts/`, discovered by `make python-tests`: 435 pass, 128 skip — the
skipped ones need the CodeQL CLI plus `cpp-queries`, which CI does not have.

Every new check was run against the broken form first and observed to fail, then against the
fix. That caught three defects in this PR's own work: an invocation block that discarded the
generator's exit status, a quality-gate line that logged every failure as "exit 0" because
`$?` inside `if ! cmd` is the negated status, and a discovery script that listed the same
database twice when given overlapping roots.

`test_shell_blocks.py` parametrizes six checks over every bash block in the skill, so its
case count tracks how much markdown the skill contains rather than how much is covered. The
fixture tests beneath it are what prove the detectors still fire.

Version: `static-analysis` 1.2.2 → 1.3.0.


GrosQuildu (commented):
there are some copilot/claude comments to be addressed first

GrosQuildu (commented):
Another observation: PR adds new workflow and lot of script, but SKILL and workflow prompts are not shortened. This is susp, I would expect that a lot of current SKILL and workflow prompts can be trimmed thanks to scripts/JS workflow

kumarak:
> Another observation: PR adds new workflow and lot of script, but SKILL and workflow prompts are not shortened. This is susp, I would expect that a lot of current SKILL and workflow prompts can be trimmed thanks to scripts/JS workflow

Thanks, @GrosQuildu. I am trimming it down further and removing some tests because CI does not actually run. They will be sitting there redundant.

kumarak:
I trimmed down on some of the changes. Most of the lines of change are from the test file. PR adds ~2203 lines to the test file, but the actual workflow changes are only ~980 lines. I can further reduce the test if it is not providing much value.

kz-tob:
need the p1 fixed at minimum and the two DB regressions in the bot comments

```
P1 — codeql-build.js hardcodes a repo-relative skill path, so every file reference it hands an agent is missing for an installed plugin
```

GrosQuildu (dismissed):
Pls run pr-review a few times on it, but in general lgtm

GrosQuildu (dismissed):

kz-tob (approved):
restamping as recent changes were just a version bump and minor fixes

PR #228, opened 2026-07-31
Convert audit-context-building to a dynamic workflow

KernelClint:
## What this does

`audit-context-building` held a fixed plan that ran the same steps over many functions and spawned subagents from prose. That is a workflow, so it is one now.

`workflows/audit-context.js` ships as `/audit-context-building:audit-context <path>` and runs three phases:

1. **Orient** — map the modules, the ways in from outside, who can reach them, and the data that persists between calls. Rank the functions that carry the most weight.
2. **Analyze** — one `function-analyzer` subagent per function. Each writes its prose to `audit-context/functions/` and returns a schema-validated record.
3. **Synthesize** — derive the rules that span several functions and write `audit-context/DOSSIER.md`.

Only the records cross back into the calling session. A subagent bound to a return schema has nowhere to put a wall of text, which is what keeps per-function prose out of the context window. The skill routes to the workflow rather than analyzing inline.

## Other changes

- `SKILL.md` said a section needed a minimum of 5 assumptions, `OUTPUT_REQUIREMENTS.md` said 3 for the same section, and the agent was pointed at both. Resolved.
- `OUTPUT_REQUIREMENTS.md` and `COMPLETENESS_CHECKLIST.md` collapsed into one `ANALYSIS_FORMAT.md`. The per-function checklist existed in four places.
- Numeric quotas removed. Minimum counts of invariants, assumptions, and applications of a technique produce padding rather than analysis.
- `commands/audit-context.md` deleted. Its `--focus` flag reached a skill that never accepted one, and its command name collided with the workflow's.
- `DOMAIN_NOTES.md` added, mapping the format across smart contracts, C and C++, decompiled firmware, and web services.
- Worked example now covers C and Solidity.
- Two README links pointed at plugins that do not exist. Fixed.
- User-facing text rewritten in plainer language.
- `SKILL.md` is 302 lines to 78.

## Tests

Four cases under `evals/`:

```bash
claude plugin eval plugins/audit-context-building --judge-model sonnet
```

Run it **by name** to get the with/without arm.

| case | target | checks |
| --- | --- | --- |
| `dispatches-not-inlines` | C | routes the work instead of analyzing in the caller's context |
| `contract-continuity` | Solidity | finds the `unchecked` branch of `_charge`; treats `feeSink.notify()` as untrusted |
| `decompiled-continuity` | Ghidra output | the copy length comes from a function not in the listing; does not claim an unnamed function "is `memcpy`" |
| `continuity-across-calls` | C | the bound sits two calls down, on one of two branches |

### What the numbers actually show

Be skeptical of these tests. Measured with `--ablation with-without`:

- **`contract-continuity` and `decompiled-continuity` both score 1.00 with no plugin loaded.** A capable model finds the `unchecked` branch, treats the unimplemented interface as untrusted, and declines to assert what an unresolved symbol does, entirely unaided. These two confirm behavior against regression. They do not demonstrate the plugin causes it.
- **`continuity-across-calls`** measured Δ +0.30 on the last run and Δ +0.07 on an earlier one. Treat it as noisy at n=2–3.
- **`dispatches-not-inlines` has not been reliably measured.** It scored Δ +0.33 on an earlier revision. On later runs the plugin arm was hitting `max_turns: 20` and being truncated before it answered, which the graders scored as a routing failure. The limit is now 40 and the case has not been re-run since.

### One known cost

The plugin arm uses roughly twice the turns of a bare agent on the same prompt, because `SKILL.md` points at three reference files that are read before any work starts. That is real overhead, not a harness artifact.

## For the reviewer

Per the Maker Week rules this needs a cold run before it merges. Two things worth doing:

1. Run `/audit-context-building:audit-context` against a real target. Check that `DOSSIER.md` and the per-function files land on disk, and that what comes back to your session is a short index rather than the analysis. That behavior is the point of the change and the eval harness cannot test it, since it invokes skills and not workflows.
2. Run `claude plugin eval plugins/audit-context-building --judge-model sonnet` and see what `dispatches-not-inlines` does at the new turn limit. That is the open question above.

`CODEOWNERS` is unchanged — @omarinuwa is still the author and I did not add myself.


KernelClint:
Correcting the Tests section above: the line calling `dispatches-not-inlines` "the case that separates from a no-plugin baseline" was based on a measurement of an earlier revision of `SKILL.md`.

A later readability pass reworded the routing section, and a re-run of the full suite shows that case now scores **worse** with the plugin than without it. The other three cases are unaffected.

I am restoring the wording that measured well and re-running that case. Will post the numbers and update the PR body when it lands. Please hold off reviewing the Tests claims until then — the rest of the diff is unaffected.

omarinuwa (approved):

kz-tob (approved):

PR #230, opened 2026-07-31
Convert spec-to-code-compliance to a dynamic workflow

KernelClint:
Maker Week verdict on `spec-to-code-compliance`: **convert to a workflow**. The skill held a fixed seven-phase
plan and drove a subagent from prose, and its Phase 3 required line-by-line YAML IR for every function in the
codebase before Phase 4 could begin. On any real target the context window is exhausted partway through, so the
first few requirements get a real check and the rest get a plausible one — and the two read identically.

## What it does now

`workflows/spec-compliance.js` extracts requirements once, then gives **each requirement its own agent** to hunt
the code with, so no context ever holds the whole behavioral model. Divergences go to two agents that did not
produce them, one re-reading the code and one re-reading the document, and anything either knocks down is
dropped. A separate agent sweeps the reverse direction for behavior no document mentions, which a
requirement-driven pass cannot reach.

SKILL.md is 113 lines, down from 357. It keeps what stays judgment: which verdicts matter, and whether a gap is
a code fix or a documentation fix.

Removed and replaced:

- `OUTPUT_REQUIREMENTS.md` — set minimum item counts ("minimum 10 Spec-IR items", "at least 3 invariants") in
  the same file as "zero speculation". A spec with fewer requirements can only meet a quota by inventing them.
  The workflow's schemas carry the structure instead.
- `COMPLETENESS_CHECKLIST.md` — a self-verification pass.
- `IR_EXAMPLES.md` — 417 lines demonstrating YAML formats the schemas now enforce. Its domain calibration is
  kept in a shorter form, below.
- `commands/spec-compliance.md` — forwarded two arguments to the skill, required naming the spec that Phase 0
  exists to discover, and would have collided with the workflow on the same slash command.

Added, following the `audit-context-building` conversion: `DOMAIN_NOTES.md` maps what counts as a specification
and where enforcement hides across contracts, C and C++, services, and decompiled firmware, plus scoping a check
against an RFC rather than a project document. `WORKED_EXAMPLE.md` carries three requirements to a verdict, one
per verdict that is easy to get wrong. `ANALYSIS_FORMAT.md` holds the on-disk format in one place.

## Bugs fixed

- SKILL.md classified undocumented behavior as `UNDOCUMENTED CODE PATH` while the other two resources called it
  `code_stronger_than_spec`, and the former was not one of the six legal `match_type` values — so following the
  skill emitted a verdict outside its own enum.
- Requirement ids came from independent per-document agents, so two documents both numbering from `REQ-01`
  collided in the report and in on-disk filenames. Confirmed real in the live run below.
- An unresolvable `agentType` failed all six requirement checks and returned nothing salvageable. It now falls
  back to the default agent with the checker's rules inlined.

## Evidence

Ran end to end against `evals/documents-contradict/fixture`: 26 requirements from two documents, 6 checked
(fan-out limit), 3 divergences found, none refuted, none unverified. It found the operator zeroing balances via
`reassign` against the spec's "MUST NOT be able to reduce a member's balance", and the Senior-tier collateral
bypass from both directions. It reported the SPEC/README fee contradiction as a documentation fix, noting the
README presents the change as deliberate. The report named all 20 unchecked requirements as unknown rather than
compliant, and declined to give the unchecked fee requirements a verdict while still reporting the contradiction
as a direct observation. 23 agents, 580k subagent tokens, 332s.

Three eval cases, run with `--ablation with-without --judge-model sonnet`:

| Case | With | Without | Δ |
|---|---|---|---|
| `routes-not-inline` | 1.00 | 0.00 | +1.00 |
| `name-is-not-evidence` | 1.00 | 1.00 | 0.00 |
| `documents-contradict` | 1.00 | 1.00 | 0.00 |

The two zeros measure the model rather than the plugin — Opus 5 finds the planted defects unaided in four turns.
Both case files say so in their own descriptions and are tagged `no-delta`; they are regression guards, not
evidence of benefit. Making them discriminate needs harder fixtures.

## For the reviewer

Two things worth knowing before reading the eval numbers. `add_dirs` resolves the fixture to a path **inside the
installed plugin**, so a no-plugin arm can read `SKILL.md` and the workflow script off disk and describe the
routing perfectly with nothing loaded — that produced a Δ −0.78 that was pure plumbing. And the Workflow tool
does not execute inside an eval run. Both are why `routes-not-inline` grades tool invocation rather than response
text.

`tool_used` counts attempted calls, not successful ones, so that case is an OR over two redundant causes:
deleting the routing section from SKILL.md still passes 2/2, and deleting the workflow while leaving SKILL.md
also passes 2/2. Removing the plugin is the only red condition. Isolating further needs a `file_exists` grader
on the report, which means running the full fan-out on every eval run.

The live run exercised the **fallback** agent path, because the plugin was installed mid-session and its agents
were not in the registry. The specialized `spec-compliance-checker` path has not been run end to end; a fresh
session would exercise it.

CI gates run locally and pass: validator self-test, version-increment against base (1.1.1 → 2.0.0), 347
references resolved, hardcoded paths, personal emails, Claude and Codex loadability.

Per the Maker Week rules this needs a partner to run it cold before merge. `/spec-to-code-compliance:spec-compliance <path>` on anything with a spec and an implementation; note the 1.x command is gone and the arguments inverted (path first, spec discovered).


KernelClint:
Addressed all eight in a3acb78. One correction and one scope reduction.

**P1 `allowed-tools` omits `Workflow`** — fixed, now `Workflow Task Read Grep Glob`. Worth flagging beyond this PR: `audit-context-building`'s SKILL.md has the identical line and the identical problem, since its primary path is also a workflow dispatch.

**P2 failed checks unrecorded** — correct, and the worse half is that `requirementsChecked: 6` was indistinguishable from a run that only selected six. `pipeline()` preserves input order, so the nulls name the failures. Now logged, passed into the report prompt as unknown-rather-than-compliant alongside the deferred cut, and returned as `checksFailed`.

**P2 fallback gated on an error string** — one correction: the pattern was not guessed. The first end-to-end run emitted `agent({agentType}): agent type 'spec-to-code-compliance:spec-compliance-checker' not found. Available agents: …`, and `/agent type .* not found/i` was written against that observed message.

The underlying point stands anyway, and it is the better argument: matching any phrasing is fragile because a reworded runtime message restores the outage. It now falls back on **any** failure of the typed dispatch, and uses the message only to decide whether to latch — a resolution failure will not fix itself, whereas a transient error should retry one requirement rather than downgrade every remaining one.

**P3 refuted divergences invisible** — fixed. The data was already reaching the report prompt; the prompt was telling it to suppress them. Now it carries a "considered and dropped" list with the reasoning, on exactly the grounds you give: one refuter is enough to drop a finding, so the refutation should be auditable rather than authoritative.

**P3 nothing verifies the writes** — fixed, minus the part I decided against. The script has no filesystem access, so it cannot confirm a write; what it can stop doing is inventing the path. `report` is now `null` with a `reportWarning` when the agent named none, instead of returning a path to a file that may not exist. I also made `analysisFile` **required** in `ALIGNMENT_SCHEMA`, so an unwritten analysis is a schema failure rather than a silent omission from a directory the report tells the reader to open — the mechanism doing the work rather than a prose instruction.

I did not add a verifier agent. It would only be able to re-attest what the schema now enforces, and the Opus 5 guidance is explicit that added verification steps compound with the model's own behavior without improving results.

**P3 `{path, spec, limit}` documented as slash-command input** — correct and a real doc bug; the object literal becomes `root`. Both READMEs and SKILL.md now say the slash command takes a path and that the object form comes from asking for the run in words.

**P4 `limit: 0`** — correct, `0 ?? x` is `0`. Clamped below as well as above.

**P4 root README table** — fixed.

Re-verified: validator self-test, version-increment against base, 347 references resolved, hardcoded paths, personal emails, Claude and Codex loadability, `plugin validate`. Plus unit checks on the three new code paths — the limit clamp, the failed-id derivation from pipeline nulls, and the no-report return shape.

Not re-run end to end after these changes; the last full run is the one in the PR description.

omarinuwa (approved):

kz-tob (approved):

PR #232, opened 2026-07-31
variant-analysis: convert the skill to a dynamic workflow 

icemonster:
Restructures the variant-analysis plugin around a skill and a
dynamic workflow; adds progressive disclosure of knowledge; adds an eval that measures whether the workflow helps.

## Skill: one reference per stage

`METHODOLOGY.md` is now four stage-scoped references. Each is loaded only by the stage that needs it:

| Stage | Reads |
|---|---|
| Root cause | `references/root-cause.md` |
| Baseline | nothing — it runs one command and reports the count |
| Sweep | `references/searching.md` |
| Triage | `references/triage.md` |
| Report | `references/reporting.md` |

A sweep agent now loads 159 lines instead of 327.
`SKILL.md` drops 142 → 76 lines and becomes the entry point.

## Workflow: `/variant-analysis:variants`

`workflows/variants.js` orchestrates the five steps across parallel subagents, one per expansion axis, looping until the sweep stops finding anything new.

The command was a stub forwarding to the skill, added in #49 before workflows existed. Skills are model-invoked, so the skill already triggers on the case the command claimed to own, and its routing advice duplicated `SKILL.md`. 

## Eval

`tests/` measures whether the workflow beats the skill alone. It fetches gradio at a
pinned SHA and injects two lookalike command-injection vulns plus a decoy. The hunt
is seeded with one vuln; finding the second is what's measured.

The variant is deliberately hard to reach: different module, different sink (`subprocess.run(shell=True)` vs `os.system`). Grepping the seed's exact sink does not find it.

**Results so far are honest but mixed.** Both modes find the injected variant. The
one real differential: the workflow found `screen_recording_utils.py:281` — a genuine
pre-existing `create_subprocess_shell` with an interpolated path in upstream gradio —
which the baseline missed. It also surfaced ~9 lower-value locations. So: one real
find, at the cost of more triage. n=1; not yet a stable result.

Five smaller synthetic codebases were tried first and showed no difference between modes. Small enough that one context handled them, so fan-out bought nothing. That negative result is why the eval moved to one large real codebase.

### Notes for reviewers regarding the tests
- `eval.sh` calls Claude and is deliberately named outside the `run_*.sh` pattern `make shell-suites` discovers.
- Findings outside ground truth are reported as `unreviewed`, not false positives.
- `work/` is gitignored; only the 79-line patch and ground truth are committed.


kz-tob:
@icemonster please review/merge the PR from @KernelClint into your branch

KernelClint (approved):

kz-tob (approved):

PR #234, opened 2026-08-02
Add a workflow for porting Semgrep rules to other languages

mathcelo:
Porting a Semgrep rule is the same phases repeated per language, so
`semgrep-rule-variant-creator` gains a workflow script rather than instructions to
re-follow each run. The script encodes the phase order, so a rule cannot be written
before the tests that specify it, and a run reports four outcomes per language:
passed, failed validation, not applicable, and unsupported by semgrep.

## What the script holds that an agent otherwise would

Four decisions have no oracle behind them, and each one fails quietly when it goes
wrong — the run still reports every language as passed.

**A NOT_APPLICABLE verdict goes to an independent refuter.** It produces no tests, no
rule, and no directory, so nothing downstream can contradict it. The refuter treats a
matching construct as necessary and not sufficient: a sink fed only by developer-set
config is not attacker-controlled, and the verdict stands.

**Validation is measured against one specific semgrep** — the version recorded when
the rule was read. Reading the verdict out of semgrep's own words only binds the agent
while the binary is fixed. An agent that could not pass its Elixir tests installed the
last OSS build shipping the Elixir parser, ran there, and reported a true
"All tests passed" for a port that is red on the semgrep the rule has to run under. A
rule semgrep skipped rather than ran is also a failure: that ends in "All tests passed"
too, over zero graded tests.

**Whether semgrep can parse the target is asked separately from whether the bug class
exists there.** Perl has no frontend, so taint no-ops under `generic` at ~100%
"parsed". Elixir's parser is Pro-only, so the rule is skipped. In both the class is
present and the rule ungradeable, which is a different finding from NOT_APPLICABLE.
This gate drops a language with no refuter behind it, so it carries the command that
settled it into the result.

**The rule travels as a path each phase reads.** Asked to repeat the file back
verbatim into a schema field, an agent HTML-escaped `<` and `>`, breaking Semgrep's
`<... ...>` deep-expression operator for every phase downstream.

Test file extensions come from an allowlist of semgrep's own language keys, including
the aliases it accepts (`sol`, `py`, `kt`, `ex`, `tf`, `golang`, `docker`), with no
fall back to an extension an agent claimed. Semgrep skips a file whose extension does
not match the rule's language and then reports a pass for the zero tests it ran.

## Tests

Three suites, all run by `make check`: the runtime contract, the deterministic
functions, and the failure paths. The last runs the real script against stubbed agents,
so the retry loop and each guard above execute. Every guard has a mutation proving it
still detects what it exists to detect — accepting a foreign-version pass, accepting a
green over a skipped rule, porting to an unparseable language, and relaying the rule as
text all turn the suite red.

## Verification

Nine end-to-end runs against real rules from `semgrep/semgrep-rules@40b8c63`, not
synthetic fixtures. Ports that passed were re-graded by running `semgrep --test` on the
artifacts directly rather than trusting the run's own report.

- The Elixir case reported `passed` before these guards and now lands in `unsupported`,
  stopping after two agents.
- Go, Bash, Rust and C++ ports go green and re-grade green.
- The refuter held six NOT_APPLICABLE verdicts (Dockerfile `ARG`/`ENV`, Perl, Ruby,
  Python twice, Solidity) and overturned one.
- `make check`, both loadability checks, and `prek run` are clean locally.

## Two notes for review

The fixes to the guards are folded into the first commit rather than isolated. The
script was untracked when they were made, so git sees one new file; separating them
would mean reconstructing a pre-fix version rather than recording real history.

The refuter's one overturn was measured by handing it a fabricated NOT_APPLICABLE for a
port that should succeed (`node-mysql-sqli` to Go). Seven attempts could not get a real
assessor to produce a wrong verdict, including one run with the assessor dropped to low
effort. The prompt was the shipping one byte-for-byte, so it shows the refuter can
overturn, not how often it will need to.


anishnaik (dismissed):

kz-tob (approved):

kz-tob (changes requested):
need to address findings from github actions run

kz-tob:
need the P2's fixed at minimum still. 
```bash
Six findings, none blocking-critical:

- **P2** `SKILL.md:30-44` — `ls DIR` prints entries, not the path, so "pass the value as printed" yields nothing usable; the script's absolute-path guard then aborts the run. `ls -d` fixes it (and the wording is pinned by a mutation test, so both need updating).
- **P2** `SKILL.md:34` — single-step `${CLAUDE_PLUGIN_ROOT}` resolution with no fallback, unlike c-review/rust-review's three-step chain; unreachable under Codex or any harness that doesn't export it.
- **P3** — the Pro-skip guard reads a "last 20 lines" window that can exclude semgrep's skip notice, reproducing the exact vacuous-green class the PR targets; `--test --json` (already used by the golden grader) is immune.
- **P3** — CI installs and asserts semgrep but never sets up node, so the whole failure-path mutation battery is one runner-image change away from silently skipping.
- **P4** — the relay checker only knows the spelling `rawYaml`; **P4** — rule id isn't slugged into the output path while the language is; **P4** — 1.0.1→1.3.0 bump size.

Verifier-integrity, agent wiring, HTML escaping, and result truncation are clean, and I said so explicitly.
Review posted (1 comment(s) created or updated since 2026-08-05T20:40:21Z).
```

anishnaik (approved):

kz-tob (approved):
Findings were addressed and rule conversion dynamic workflow was tested locally

PR #235, opened 2026-08-03
Rework property-based-testing skill and add eval suites.

gsutherland-trailofbits:
Reworks the property-based-testing skill and adds two eval suites.

`evals-extra/` holds shell harnesses run by hand: trigger rate against labelled queries, and whether a generated suite catches a planted defect. Both ship a free `--self-test` that runs in `make check`.

`evals/` holds `claude plugin eval` cases run under ablation, so the number reported is Δ against the unaided model. Measured on opus, 3 runs per arm:

  | case | fires | with | without | Δ |
  |---|---|---|---|---|
  | 02-neg-cargo-fuzz-coverage | no | 1.00 | 1.00 | 0.00 |
  | 03-dependency-is-users-call | yes | 0.50 | 0.00 | +0.50 |

02 guards over-triggering; the skill did not fire in any with-plugin run. 03 measures whether a new test dependency is left to the user. Unaided opus wrote a requirements.txt in 3 of 3 runs, and 2 of 3 with the skill loaded, so the behaviour is improved and not fixed. The grader carrying that Δ is a regex over the created-files list rather than an LLM judge.


kz-tob:
resolve conflicts and need a partner to review/test

elopez:
@claude re-review all concerns and resolve conversations if addressed already

elopez (dismissed):
I tested this locally with a sample solidity project and built some Echidna tests, it worked fine there. I pushed some general fixes from local testing with Claude, feel free to revert/rework them if you don't agree with them.

Having said that, one concern that Claude brought up and I thought was valid:
> **`refactoring.md` was deleted and its content survives nowhere.** Seven patterns for making untestable code testable: extract the pure core, add the missing inverse, inject dependencies, return instead of mutate, structured + render. The new SKILL.md ends that path at "Code with no such shape gets example tests; saying so is a valid outcome" — a dead end where the old skill had an answer.
>
> `evals/03`'s own fixture is that case: `send_welcome_email` is impure SMTP, and the property-worthy part is building the `EmailMessage`, separable from `s.send_message`. Deleting `design.md` (a Phase 1–5 prose workflow, which AGENTS.md says belongs in a script) and `strategies.md` (per-language syntax the model already has) both look right to me; this one reads like collateral. Either restore a trimmed version or add it to the deliberate-cuts list, because right now it is the one deletion neither README accounts for.

gsutherland-trailofbits:
@elopez Restored as suggested. Good to go.

kz-tob (approved):

kz-tob:
The CLA is failing, @elopez @gsutherland-trailofbits [recheck your sign status here](https://cla-assistant.io/check/trailofbits/skills?pullRequest=235)

elopez:
@kz-tob CLA appears to be green now

PR #237, opened 2026-08-03
variant-analysis: fix problems found by testing #232 before release

KernelClint:
This sits on top of #232. Merging it folds these fixes into that PR, which then goes to `main` as one change.

**What I did:** ran #232's plugin the way a user would, found bugs, fixed them. Nothing here is a redesign. It is #232 plus the things that broke when it was actually used.

## The short version

| Problem | Why it mattered | Fix |
|---|---|---|
| The workflow died instantly if Claude described the bug in a sentence instead of a JSON object | Whole run lost, no output, no clue why | Accept both |
| A codebase path with a space broke the search command | Silent wrong results | Quote it properly |
| 25 subagents ran against a 5-file project | Slow and expensive for no benefit | Small projects now use 7 |
| The scorer marked a *correct* report as wrong | The test failed good runs | Score against a function's real line range, not "within 30 lines" |
| The scorer's main path never ran on real reports, and nothing said so | Scores looked better than they were | The report now writes the format the scorer reads |
| A documented option (`--strict-decoy`) didn't exist | Following the README got you an error | Wired it up |
| Two test graders could never fail | They inflated the score and measured nothing | Deleted |
| Nothing checked the workflow script for syntax errors | A typo would only surface in a paid test run | Free syntax check added to CI |

Plus smaller things: a confusing error message, a path-parsing bug, and `AGENTS.md` didn't document where dynamic workflows go — this is the repo's first one.

## One claim in #232 was wrong

#232 says the skill never activates on its own, based on 40+ runs. I re-tested it 9 times and found the real story:

| How the user asks | Is there a codebase to search? | Activates? |
|---|---|---|
| "found a bug in `x.py`, are there others?" | yes | 2 of 3 |
| Wording closest to the skill's own description | yes | 3 of 3 |
| Code pasted into the chat, no files | no | 0 of 3 |

The skill activates fine. It just doesn't activate on pasted code — and every test case in #232 pastes code. With no files to search, there's nothing for it to do, so skipping it is arguably correct.

This matters for two reasons. It means deleting the old `/variants` command was the right call (Claude finds the workflow on its own, 4 times out of 5). And it means the README was overselling the skill, so I rewrote that section to say what actually happens.

## How I know the fixes work

Same workflow, same test project, before and after:

| | before | after |
|---|---|---|
| subagents used | 25 | 7 |
| scorer used its main path | no | yes |
| flagged a safe site as a bug | yes (wrong) | no |
| test result | FAIL | PASS |
| found both planted bugs | yes | yes |

I also broke each new test on purpose to confirm it fails when it should. Grader tests went from 13 to 18, aggregator tests 6 to 7. `ruff`, `shellcheck`, `shfmt`, the offline test suite, and `make validate` are all clean.

## Still not fixed

The test proves the workflow doesn't break. It does **not** prove the workflow beats the plain skill — both find the planted bug, so a tie passes.

Fixing that means giving the test cases a real codebase on disk. That would also solve the skill-activation gap above and the fact that the test scores are maxed out. It's the most valuable work left on this plugin, and it's too big to attach here.

## Reviewer notes

- **CI looks thin on purpose.** `lint.yml` and `validate.yml` only run for PRs targeting `main`, so a stacked PR gets neither. I ran their exact commands locally; they'll run for real once #232 merges.
- Version is 2.0.1. Squash it to 2.0.0 if you'd rather ship one version number.
- The comment thread below has the detailed back-and-forth with the automated reviewer, including one point of its I pushed back on.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -214,7 +214,7 @@
     },
     {
       "name": "variant-analysis",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "description": "Find similar vulnerabilities and bugs across codebases using pattern-based analysis",
       "author": {
         "name": "Axel Mierczuk"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,17 +68,29 @@ plugins/
       plugin.json         # Plugin metadata (name, version, description, author)
     commands/             # Optional: slash commands
     agents/               # Optional: autonomous agents
+    workflows/            # Optional: dynamic workflows (*.js), ships as /<plugin>:<workflow>
+    evals/                # Optional: `claude plugin eval` cases + graders
     skills/               # Optional: knowledge/guidance
       <skill-name>/
         SKILL.md          # Entry point with frontmatter
         references/       # Optional: detailed docs
-        workflows/        # Optional: step-by-step guides
+        workflows/        # Optional: step-by-step guides (prose, not scripts)
         scripts/          # Optional: utility scripts
     hooks/                # Optional: event hooks
+    tests/                # Optional: run_*.sh suites (CI runs these — keep them free)
     README.md             # Plugin documentation
 ```
 
 **Important**: Component directories (`skills/`, `commands/`, `agents/`, `hooks/`) must be at the plugin root, NOT inside `.claude-plugin/`. Only `plugin.json` belongs in `.claude-plugin/`.
+
+**Two different `workflows/`.** A **dynamic workflow** is a JavaScript file at the *plugin
+root* under `workflows/`; it exports a `meta` object, orchestrates subagents in code, and
+ships through the marketplace as `/<plugin-name>:<workflow-name>` (from `meta.name`, not the
+filename). The older `skills/<skill>/workflows/` holds prose step-by-step guides. If a
+SKILL.md has "Phase 1", "for each finding", or "repeat until" sections, the plan belongs in
+a script at the plugin root, not in prose. See `plugins/variant-analysis/workflows/` for a
+worked example, and note that `${CLAUDE_PLUGIN_ROOT}` is **not** available inside a workflow
+script — it is not a hook, MCP, or LSP subprocess.
 
 ### Frontmatter
 
--- a/plugins/variant-analysis/.claude-plugin/plugin.json
+++ b/plugins/variant-analysis/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "variant-analysis",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "Find similar vulnerabilities and bugs across codebases using pattern-based analysis",
   "author": {
     "name": "Axel Mierczuk"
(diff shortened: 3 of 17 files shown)


KernelClint:
**In plain terms:** the automated review found 6 real bugs in my fixes and I fixed all 6. The worst one: a report could claim a safe site was vulnerable, and the scorer would credit it for *correctly ruling that site out*. One of the review's points was based on a wrong reading of the code, so I left that behaviour alone and explained why below.

Detail follows.

---

Worked through the automated review. Six of its ten points were real and are fixed in 2b9b856; one premise was wrong and I've left that behaviour as-is.

**Fixed**

- **The line-less-claim gap (P2).** Correct, and the worst of the batch — a claim on the decoy's file with no line stayed out of `decoy_reported_as_real` (right), matched `known` permissively so it vanished from `unreviewed_findings`, and then *satisfied* `decoy_examined_and_ruled_out`. A run got credit for ruling out the site it had just listed under Findings. Now surfaced as `decoy_claimed_without_line`, kept in `unreviewed_findings`, and it blocks the ruled-out credit without counting as a false positive. My earlier self-test asserted the wrong behaviour here; it now asserts the right one.
- **Silently dropped axes (P2).** Also correct, and it reproduced: the re-run's report now contains a `## Never Swept Axes` section reading *"Five of seven planned axes were not swept in this run (2/7 covered)"*. Before the fix that was only in the live log. Swept/total and the unswept axis names now reach both the report prompt and the return value.
- **`loose` had no test (P3).** The labels are constants in `score.py` now, pinned by a self-test, plus a `summarize.py` row asserting `permissive == 1`. Confirmed the rename mutation fails the suite.
- **Nothing required a span (P3).** `verify_fixtures.py` now requires one on every entry and validates the range, so a dropped span can't silently revert the grader to a proximity window.
- **`source_file_count` counted everything (P3).** Right — `rg --files` includes assets and fixtures, so a 25-file project behind 300 fixtures reported 325 and missed its own floor. Prompt and schema now ask for source files only, and the README says what's counted.
- **Minor:** `.mjs` copy for `node --check` (the Node-version point is fair, `lint.yml` pins none); self-test fixture no longer uses a span that excludes its own anchor; dead `!out.bug` clause dropped; third-person description per `AGENTS.md:102`.

**Not changed — the premise is wrong**

P3 on spans being "the sink's neighbourhood, not the enclosing function." They are the enclosing function. From `gradio-vulns.patch`: `def archive_flagged_media` is line 348 and its `return archive_path` is 352, span `[348, 352]`; `def extract_video_thumbnail` is 1103 and its `return output_path` is 1111, span `[1103, 1111]`. The suggested failing citation, `processing_utils.py:1098`, is five lines *above* the def — inside `video_is_playable`, a different function — so declining to credit it is the intended behaviour, not a regression.

The narrow real case underneath it is a decorator or overload stub immediately above a `def`, which an exact span does exclude. Added `RECALL_PAD = 3`, applied to recall only. The safe site gets no slack, since padding it walks straight back into the conflation the spans exist to fix. Both directions are covered by a new self-test.

**Re-verified end to end** (same fixture, after all of the above): 7 agents, `extraction_mode: location-fields`, recall 2/2, `decoy_reported_as_real: false`, verdict PASS. `score.py` self-test 16 → 18, `summarize.py` 6 → 7; `ruff`, `shellcheck --severity=warning`, `shfmt -i 2 -ci`, `run_fixtures.sh`, and `make validate` all clean.

icemonster (approved):
lgtm

PR #238, opened 2026-08-04
c-review rewrite to workflows + do measurements

GrosQuildu:
Rewrote c-review from per-bug-class agent sweep on whole codebase in two steps:
1. agents working on the same questions (return-values, alloc-lifetime, ...) where every is given a part of the codebase
2. sweep of agents focused on a single bug class that did not produce any findings in step 1

This make plugin's agents much less constrained than before, and put more focus on `location x bug classes`  coverage rather than only `bug classes'.

Removed fp-judge as it seems redundant: we have another skill for that.

Kept deduplication, made it more deterministic.

The measurements show that c-review gives only a small improvement over simple "fanout" (while spending more tokens), but the skill is structurally promising: more focused agents, and checks on the actual coverage.
I think the fanout vs c-review numbers would converge after a few re-executions on current evals. Measurements on larger codebase is needed to determine if the skill is really useful.

The benchmark code is in this PR but is removed from final version.

## Measurements

c-review v2.0.0 was measured against two baselines on two corpora. The benchmark harness is
not in this PR — it carries sealed ground truth that must not be public — so everything
needed to judge the numbers is below.

**How it works.** Two C corpora with bugs **we injected ourselves**, so no CVE database
holds the answers. Each arm reviews the same tree; a grader matches findings against private
ground truth by site. Every cell ran in a hermetic container (1 plugin, 18 skills, no host
memory, network blocked). An anti-cheat pass parses transcripts for real tool calls, and a
cell that consulted an upstream source or reached for the answer key is **excluded**, not
annotated. Every cell below is VALID. Tokens are `tokens_fresh` (input + output + cache
creation).

**Arms.** `bare` = one plain prompt, one agent. `fanout` = 13 generic agents over
contiguous, disjoint code regions — no taxonomy, no judges, no orchestration, each
inventing its own bug labels. It partitions by location exactly as c-review does, which
is what makes it the sharp baseline: it isolates what the orchestration adds over a
naive split. `c-review` = the plugin as shipped.

### sigil — 17 injected bugs, 901 lines

| arm | recall | EASY | MEDIUM | HARD | agents | tokens | wall |
|---|---|---|---|---|---|---|---|
| bare | 11/17 | 5/5 | 6/9 | 0/3 | 1 | 0.18M | 8 min |
| fanout | 14/17 | 4/5 | 7/9 | 3/3 | 13 | 1.06M | 5 min |
| **c-review v2.0.0** | **16/17** | 5/5 | 9/9 | 2/3 | 8 | 2.10M | 35 min |

### packetloom — 30 injected bugs, 3,434 lines

| arm | recall | EASY | MEDIUM | HARD | agents | tokens | wall |
|---|---|---|---|---|---|---|---|
| bare | 18/30 | 5/11 | 7/9 | 6/10 | 1 | 0.36M | 10 min |
| fanout | **24/30** | 7/11 | 8/9 | 9/10 | 13 | 1.39M | 10 min |
| **c-review v2.0.0** | **24/30** | 6/11 | 8/9 | **10/10** | 8 | 2.71M | 45 min |

> **One cell per arm, and the noise floor is 1–2 bugs out of 17** — two runs of an identical
> `bare` cell previously scored 11 and 10. A 2-bug gap is suggestive, not conclusive.

### What this says

1. **Both structured arms beat a single prompt, and the hard tier is why.** `bare` scored
   **0/3** on sigil's HARD bugs and 6/10 on packetloom's; c-review scored 2/3 and **10/10**.
   That gap is the justification for spending more than one agent.
3. **c-review is +2 on sigil and level on packetloom.** Read that as "at least as good as
   the strongest baseline", not as a win.
4. **fanout is about half the cost and a quarter of the wall time** — 1.06M vs 2.10M, 1.39M
   vs 2.71M, at 5–10 minutes against 35–45. c-review's phases are sequential by design;
   fanout's agents all run at once. **There is no cost argument for c-review.**
5. **The argument is verified coverage.** `fanout` also owns every line exactly once, so
   "every line is assigned" is not the differentiator. c-review's partition is generated
   from a tree-sitter parse at function granularity, and a gate afterwards re-derives the
   sites from source and checks each agent's account against them. `fanout` *assigns*
   coverage; c-review *measures* it, and reports the shortfall as a number. If you don't
   need that, fanout is the cheaper tool.


GrosQuildu:
## How the measurements were protected

Two threats, handled separately. Both fired for real during these runs.

**Context pollution.** Every cell runs in a hermetic container — 1 plugin, 18 skills,
container-local memory, network blocked by a `PreToolUse` guard — and isolation is verified
per cell from the session's own `"subtype":"init"` record rather than assumed. This matters
because host runs were badly polluted: ~48 skills reached every agent, *including other
security-review skills and c-review itself*, plus a foreign `SessionStart` hook that
executed four times and injected its stdout into the context. The effect was real and
asymmetric — hermetic `fanout` scores **14/17** where the polluted host gave 16 and 15,
while `bare` was unmoved (11, 10 → 11). Any baseline figure predating the container is
superseded.

**Agents reading the answers.** Ground truth is encrypted at rest, not just hidden, and
`bench.py plan` refuses to emit packets unless every corpus is sealed. Bugs are injected at
sites we chose, so no CVE database holds them, and both corpora used here were authored for
the harness with no upstream to look up. (De-identification alone is not sufficient — a
de-identified zlib corpus is marked *do not use* because the model recognised it from
memory and said so; blocking the network did not help.) An anti-cheat pass parses real
`tool_use` blocks from each transcript — not substring matches, since `WebFetch` appears in
almost every transcript as a tool *definition* — and a cell that used an oracle is
**excluded from the comparison, not annotated**.

**Both controls fired.** A `fanout` agent ran `find /corpus -iname "recipe.json"`. The seal
held — no recipe in the tree, the probe returned nothing — and the gate excluded the cell
anyway; it was re-run rather than argued with. Separately, the corpus turned out to name its
own answer file in a source comment. That was checked rather than assumed harmless: the file
it discusses holds three injected bugs, and all three arms found all three, so recall was
unaffected. The cost was one excluded cell.

**Known gaps**, for completeness: that corpus comment should be removed; the detector
excludes on the *attempt* while the harness's stated policy says attempted-and-denied is not
disqualifying (implementation and policy disagree, and the gate was not overridden); and
every arm is told it is a benchmark by its packet, so subtler behavioural effects cannot be
ruled out — though that applies uniformly across arms.

GrosQuildu:
Ignore CLA please, I committed from not-existing email.

GrosQuildu:
Replaced by https://github.com/trailofbits/skills/pull/257

PR #240, opened 2026-08-06
Add an eval suite for entry-point-analyzer

mkzung:
`entry-point-analyzer` had no evals. Two Solidity cases.

Ground truth is the solc ABI rather than my reading of the source: an entry point is an ABI member that is neither `view` nor `pure`. That caught a fixture of mine that did not compile, and it is what the counts below come from.

**`inherited-and-unnamed`** covers the entry points that are not in the file named after the contract. Four of the eleven on `StakingVault` are declared in `AccessBase` and `Pausable`, and `receive()` has no name to grep for. Against those, six of the nine read-only members are compiler-generated getters, which read like functions in a report but cannot change state. Four scored graders: the inherited four, `receive`, the read-only exclusions, and the access classification. `fallback()` reverts, so it is outside every gate.

**`no-state-changing-surface`** is the control. `FeeQuoter` has none, though the constructor writes `tierOf` in a loop and two view functions read it back. Its two graders fail on opposite mistakes, one on padding the report with read-only members and one on never stating the conclusion, so neither is satisfied by an empty answer.

`case.yaml` rather than `prompt.md` because the fixtures span three files and `context.add_dirs` is how `audit-context-building` and `spec-to-code-compliance` already do that. Version 1.0.1 to 1.1.0 in both places.

## What I have not done

**No ablation.** I have not run these against a judge, with or without the plugin, so I cannot tell you the suite detects a broken skill. After reading #229 I think that is the bar, and this does not clear it yet. Ground truth is verified by the compiler and the rubrics are checked by hand, which is a different and weaker claim. Happy to run it if you can point me at how you produce the numbers in #229, or to hold this until I can.

One thing worth flagging either way: the skill routes on a description that says smart contracts, and these prompts are phrased as an auditor scoping work rather than as a request for entry point analysis. Both cases carry a display-only `skill-fired` grader so the trigger rate is visible.

`make self-test`, `make lint` and `make validate` pass. Note that `validate_reference_links` skips any path with `evals` in it, so validate does not actually cover this directory. `make python-tests` fails the same way on a clean checkout, in `constant-time-analysis`.

Written with Claude Code, reviewed and verified by me.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -115,7 +115,7 @@
     },
     {
       "name": "entry-point-analyzer",
-      "version": "1.0.1",
+      "version": "1.1.0",
       "description": "Analyzes smart contract codebases to identify state-changing entry points for security auditing. Detects externally callable functions that modify state, categorizes them by access level, and generates structured audit reports.",
       "author": {
         "name": "Nicolas Donboly",
--- a/plugins/entry-point-analyzer/.claude-plugin/plugin.json
+++ b/plugins/entry-point-analyzer/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "entry-point-analyzer",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "description": "Analyzes smart contract codebases to identify state-changing entry points for security auditing. Detects externally callable functions that modify state, categorizes them by access level, and generates structured audit reports.",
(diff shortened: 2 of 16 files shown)


kz-tob:
Thanks for the contribution.

We're not accepting eval-only external contributions without demonstrated uplift — suites need scoring from an ablation run attached, showing separation from an unaided baseline. Closing this one.

PR #243, opened 2026-08-07
fp-check - merge with concept-prover and online-triage, add evals

GrosQuildu:
Rewrites fp-check into workflows, merges it with concept-prover and online-triage (as optional steps); adds tests and evals.

Based on measurements it works as good as the old fp-check and concept-prover run together, but saves tokens and provides single unified skill.

## What it does

Decides whether a suspected security bug is real, and says why. You hand it a
finding; it returns **TRUE POSITIVE**, **FALSE POSITIVE** or **NEEDS MORE INFO**
with the evidence behind the answer.

Three stages, in order:

| Stage | When | What it does |
|---|---|---|
| **0 Batch** | optional | derives the shared context once, runs Stage 1 per finding with it, accounts for every finding by id, and then checks the pairs that are only exploitable together |
| **1 Static** | always | traces the path from entry point to sink, one agent per validation layer, then checks recovery, upstream fixes, impact and severity, and finally six gates |
| **2 Online** | on request | the project's published policy, bounty scope, past reports and duplicates. Refuses to answer offline rather than guessing from memory |
| **3 PoC** | on request | builds a real exploit, runs it, then five agents that did not build it try to reject it |

Stage 1 always reaches a verdict on its own. Stages 2 and 3 can only narrow or
correct it.

**The design point:** every gate is a pure function in the workflow script, not an
instruction an agent is asked to follow. The script returns a status you cannot
argue with. That distinction is the whole reason the plugin exists — the same
rules, written as prose in a prompt, were not obeyed.

Two examples of what that buys: a finding already fixed upstream is retracted and
must cite a commit, and an integration or external root cause is capped at Medium
by arithmetic rather than by asking the model to be modest.

## Stop Hooks Remove
The two Stop/SubagentStop prompt hooks were dropped rather than fixed because the completeness check they performed is now enforced deterministically by the workflow gates, while every observation we have of the hooks themselves is negative — invalid JSON on every session end, firing in sessions that never used the plugin, blocking correct standard-path output.

## What was measured

**Question:** does loading the plugin produce better security conclusions than the
same prompt with no plugin at all?

**Method:** `claude plugin eval` with an ablation baseline — the identical prompt
is run twice, once with the plugin loaded and once with no plugin at all, and the
difference between the two scores is the delta. The two configurations are called
the **with-arm** and the **no-plugin arm**.

Because a language model gives a different answer each time, every case is run
**3 times per arm** — that is what *n=3* means. So: 7 cases × 3 runs × 2 arms =
**42 runs**, and a case's score is the mean of its 3 runs. A case scores 1.000 only
if all three runs got it right.

Each case scaffolds a small realistic codebase and reports a finding the way a user
would; some reports are correct, some are wrong in a specific way. Every case is
graded by an LLM judge reading the final answer, **plus** at least one deterministic
regex or tool-use check — never an LLM grader alone, because it will happily pass a
run that *described* doing the work instead of doing it. Both arms and the judge
run on Sonnet.

### The seven cases

Each one is a finding reported to the tool, and each has a specific right answer:

| Case | The report | The right answer |
|---|---|---|
| `integration-cap` | a real ledger bug, reported as Critical | real, but the attacker does not control the upstream value, so severity caps at Medium |
| `blocked-attack-path` | SQL built by string concatenation | false positive — two validators reject the metacharacters injection needs |
| `already-fixed` | a timing oracle reported against v1.4.0 | retracted — the fix landed one layer up in the caller; only git history shows it |
| `dead-route` | shell injection in a handler | false positive — no route in the routing table reaches that handler |
| `inflated-impact` | a panic claimed as a remote process crash | real but over-stated — Go's `net/http` recovers per connection, so one connection drops and the server keeps serving |
| `wrong-parameter` | `shell=True` and attacker input in one file | false positive — there is no path between the two; the report joined the wrong sink to the wrong source |
| `should-not-fire` | *"explain what this function does, I am writing API docs"* | do nothing — this is not a security request, and the tool must not activate |

### Which stages were measured

**Only Stage 1 is measured across the board.** All seven prompts pin *"work
offline"*, so:

| Stage | Coverage |
|---|---|
| **1 Static** | measured on all 7 cases, both arms — this is what +0.281 describes |
| **3 PoC** | six prompts ask for a PoC, but Stage 3 only runs on a `TRUE_POSITIVE`, and only `integration-cap` reaches one. So it is exercised by **one case** |
| **2 Online** | **never measured.** Its one case does not discriminate — a plain session scores 1.000 on it — and it needs a suite built on real public advisories, which does not exist yet |

So this is strong evidence for the static path, thin evidence for the PoC path, and
no evidence for the online path.

## Results

**v2.0.0 — 42/42 runs clean, $29.47, 2h26m**

| | with plugin | no plugin | delta |
|---|---|---|---|
| Mean over 7 cases | **1.000** | 0.719 | **+0.281** |

Per case:

| Case | with | without | delta |
|---|---|---|---|
| integration-cap | 1.000 | 0.333 | **+0.667** |
| blocked-attack-path | 1.000 | 0.400 | **+0.600** |
| already-fixed | 1.000 | 0.500 | **+0.500** |
| dead-route | 1.000 | 0.800 | +0.200 |
| inflated-impact | 1.000 | 1.000 | 0.000 |
| should-not-fire | 1.000 | 1.000 | 0.000 |
| wrong-parameter | 1.000 | 1.000 | 0.000 |

**Free test layers:** 358 node, 382 pytest, 44 bats, validator clean. Mutation gate
**146 mutations run, 0 survived** — each covered behaviour is broken in a sandbox
and the suite is required to go red, so the tests are known to test something.

## Conclusions

**It works, and the effect is concentrated.** Four of seven cases carry the whole
delta; three contribute nothing because a plain session already scores 1.000 on
them. The wins are exactly where gates are enforced in code: rejecting an
unreachable sink, retracting an already-fixed bug, capping an inflated severity.

**Judge progress by the with-arm score, not by the delta.** The delta is a
subtraction: `with − without`. Both halves move between sweeps, so the delta can
rise because the plugin improved *or* because the no-plugin baseline happened to
score worse that day — and you cannot tell which from the delta alone.

Across the four sweeps the no-plugin baseline scored 0.656, 0.700, 0.652, 0.719.
Nothing about the plugin changed those; that spread is just run-to-run variation at
3 runs per case. The with-arm over the same four sweeps went 0.868 → 0.908 → 0.919
→ **1.000** — that one only moves when the plugin changes, so it is the number that
actually tracks progress. It is also a more modest claim than the deltas suggest.

**The suite is now saturated, and that is the binding limit.** All 21 with-arm runs
scored exactly 1.000. There is no headroom left to measure an improvement in, so
the next change to this plugin cannot be evaluated by this suite — it will report
roughly +0.28 whatever the change does. A harder case is owed before another sweep
is worth paying for.


GrosQuildu:
Ignore CLA please, I committed from not existing email.

ahpaleus:
I am on it; fixing and cleaning it up

PR #247, opened 2026-08-13
Feat/add cardano

Kammerlo:
## Summary

Adds `cardano-vulnerability-scanner` to `building-secure-contracts`, covering the extended UTxO (eUTxO) model used by Cardano. The plugin ships scanners for six chains today; Cardano is the notable absence, and it is the one whose vulnerability classes generalise least from the others.

Nothing in the Algorand, Cairo, Cosmos, Solana, Substrate or TON taxonomies transfers. A Cardano validator is a pure predicate over the whole transaction — it cannot act, only reject — so there is no reentrancy, no `msg.sender`, and no
contract-held state. Almost every real finding is a transaction field the validator failed to constrain, which is a shape none of the existing scanners look for.

Covers Aiken, Plutus/PlutusTx, Plutarch and OpShin.

## Design: defers upstream rather than duplicating

The Cardano Foundation maintains [cardano-dev-skills](https://github.com/cardano-foundation/cardano-dev-skills), a Claude Code plugin bundling current Cardano documentation and a 26-category eUTxO checklist. It exists because Cardano training data and reference material drift quickly across ledger eras.

Copying that checklist here would reintroduce the drift problem it solves, so this skill resolves it at runtime instead and treats it as the primary taxonomy when present:

```bash
find ~/.claude/plugins ~/.codex 2>/dev/null \
  -path '*/skills/review-contract/references/vulnerability-checklist.md' -print -quit
```

If it resolves, the scanner works from the upstream 26 categories. If not, it falls back to the 17 patterns in this PR **and is instructed to say so in the report** the fallback is narrower, and a reader who does not know the scan ran degraded will
read a short findings list as good news.

This follows the value-add guidance in `AGENTS.md` ("teach when and how to look things up" rather than dumping reference material) and the existing soft-dependency convention in the plugin — the Algorand scanner treats Tealer the same way.

**Disclosure:** I work at the Cardano Foundation, which maintains cardano-dev-skills. The delegation is there because that project tracks protocol releases and this file cannot, but you should weigh the recommendation knowing where it comes from. The scanner is fully functional without it.

## Contents

```
skills/cardano-vulnerability-scanner/
├── SKILL.md                              348 lines
├── resources/VULNERABILITY_PATTERNS.md   779 lines, 17 patterns
├── agents/openai.yaml
└── assets/trail-of-bits-mark.svg
```

17 patterns with vulnerable/fixed Aiken pairs, drawn from published Cardano audit work (Tweag's five-year retrospective, Vacuumlabs' *Cardano Vulnerabilities* series, MLabs and Anastasia Labs reports): double satisfaction, datum hijacking, missing signer checks, value preservation, unrestricted minting, staking credential checks, withdraw-zero bypass, state thread token mismanagement, oracle authenticity, and others.

Six further upstream classes are named with a reason they are not detailed (deployment-time rather than source-visible, subsumed elsewhere, or efficiency rather than security), so all 26 upstream categories are accounted for.


smoelius:
Hi, @Kammerlo. Thank you very much for your submission! This clearly took a lot of thought and effort to put together.

You're right that Cardano is the notable absence from our skills. But that is (unfortunately) because we don't do much Cardano work at Trail of Bits. And for that reason, I worry that this skill could rot, and we wouldn't have the expertise to fix it.

`cardano-foundation/cardano-dev-skills` seems like the better home: it's already the source of truth you're deferring to, and it has maintainers who can keep the patterns current.

Issue #233, opened 2026-08-01
How we use /goal to find bugs in Patch the Planet - $goal-prompt


colll78:
In https://blog.trailofbits.com/2026/07/28/how-we-use-goal-to-find-bugs-in-patch-the-planet/

You reference the skill `$goal-prompt`, and suggest the reader to use this skill to have the LLMs define their own goal. However, this skill doesn't appear in any repository. Is it possible that you guys could introduce it to this repository, or provide some guidance on how to reproduce this skill / how it helps facilitate the goal definition?

GrosQuildu:
Imported (with improvements) in https://github.com/trailofbits/skills/pull/248
Thanks for noticing.

PR #248, opened 2026-08-14
Add goal-prompt plugin for copy-ready /goal commands

GrosQuildu:
Migrated from internal repo and improved with our most recent data and by eval measurements.

Skill to turn a task description into a copy-paste-ready `/goal` command in Claude Code or Codex.

Solves https://github.com/trailofbits/skills/issues/233

## Evals

**Setup.** 3 cases, each the same bare prompt ("Improve this goal: … Give me the final `/goal`
command to paste.") run with and without the plugin (`claude plugin eval --ablation with-without`),
2 runs per arm per case, sonnet judge, fixtures generated inside the eval scaffold (`--scaffold`).
Graders judge the returned artifact: single-line format, grounded check command, stop clause,
no invented success criteria, easy-out closure. Two measurement invocations
(A and B below), so 4 runs per arm per case total.

## Final results (isolated scaffold, contamination check clean in both invocations)

The plugin arm passed every grader in every run (12/12). Baseline passed no case in either
invocation (0/12).

| Case | with plugin | without (A) | without (B) | pooled without | pooled delta |
| --- | --- | --- | --- | --- | --- |
| grounded-migration | 1.00 (4/4 pass) | 0.40 | 0.20 | 0.30 | **+0.70** |
| ungrounded-vague | 1.00 (4/4 pass) | 0.25 | 0.375 | 0.31 | **+0.69** |
| easy-out-closed | 1.00 (4/4 pass) | 0.29 | 0.86 | 0.57 | **+0.43** |
| **mean** | **1.00** | **0.31** | **0.48** | **0.40** | **+0.60** |

**What the baselines got wrong without the skill:** multiline rewrites that break on paste,
no stop bound (the deterministic regex grader alone fails 9/12 baseline runs), "tests pass"
unattached to any real command, grep-only success conditions satisfiable by deleting the
callers, and fabricated metrics/benchmarks for the ungroundable "make checkout faster" case.

**Run-to-run variance.** easy-out-closed is the noisiest case: in inv
did pair the fixture's real test suite (the LLM grader passed) and failed only on the missing
stop clause, lifting its arm score from 0.29 to 0.86. The other two c
graders in both invocations. Expect the easy-out delta to swing; the
grounding deltas are stable.

## Measurement pitfalls found and fixed along the way

  1. **Judge tier.** The default haiku judge could not follow scoped grading instructions
     ("judge only the line inside the fenced block") and failed correct answers for content in
     the accompanying `Missing:` list. Fixed by scoping the grader text explicitly and pinning
     `--judge-model sonnet`.
  2. **Baseline contamination.** With fixtures referenced in-repo (`context.add_dirs`), every
     agent received an absolute path into the repository — one directory walk from SKILL.md and
     the graders. Baseline responses reproduced SKILL.md sentences verbatim ("scope to read
     first", "terminates on the wrong contract") and cited the plugin's path; contaminated
     baselines scored up to 1.00 by imitation, collapsing the mean delta to +0.04. Fixed by
     generating fixtures inside the temp scaffold (`scaffold.sh` per case) — a kept-temp trace
     audit shows zero repo paths in baseline runs — and by gating every run with
     `check_contamination.py`, which fails on plugin paths or verbatim SKILL.md phrases in any
     baseline response (and fails when it has nothing to inspect).
     `check_contamination.py`, which fails on plugin paths or verbatim SKILL.md phrases in any
     baseline response (and fails when it has nothing to inspect).


dguido:
Reviewed against the bar we're now holding PRs to. This is the best-formed new plugin in the queue — three walnut-format eval cases from day one puts it ahead of 33 of the 40 plugins already merged, and `check_contamination.py` is one of the few checkers in this repo that states and implements the zero-item rule in its own docstring. Two things should be fixed before it merges, and one is worth a look.

**P1 — the skill's only operative command is refused by our own plugin.** `SKILL.md:46` says:

> Run `python3 {baseDir}/scripts/format_goal_prompt.py --fenced` on the draft

`modern-python` ships PATH shims that intercept a bare `python3 <script>` and exit 1. Verified just now against this branch:

```
$ echo "do the thing or stop after 20 turns" | python3 .../format_goal_prompt.py
ERROR: Use `uv run python .../format_goal_prompt.py` instead of `python3 .../format_goal_prompt.py`
```

So for any user who has `modern-python` installed, the Format step — the step the whole skill builds toward — fails. Note this is **not** fixed by #255: that narrowing deliberately keeps bare script invocations intercepted, because running a script against project dependencies is precisely what `uv run` exists to replace. `AGENTS.md` already prescribes the working form:

> **Use `uv run`** — `uv run {baseDir}/scripts/process.py input.pdf`

I tested `uv run <script>`, `uv run --no-project <script>` and `uv run --no-project python <script>`; all three pass the shim and produce correct output. Plain `uv run {baseDir}/scripts/format_goal_prompt.py --fenced` matches the house example. Same fix needed at `evals/README.md:10`, which invokes `check_contamination.py` the same way.

**P2 — one contamination marker can never fire.** `check_contamination.py:24` lists `"scope to read first"`, but `SKILL.md:16` writes it as `**Scope to read first**`, and `m in text` is case-sensitive. So the one marker aimed at a distinctive SKILL.md heading is dead; a baseline that copied that phrase verbatim off disk passes clean. The other four match exactly (`plugins/goal-prompt` is a path marker and correctly absent from SKILL.md). Casefold both sides, or fix the marker's capitalisation.

**P3 — the result shape is assumed, not validated.** `baseline_texts()` reads `cases[].arms.without[].graders[].evidence`, and `test_check_contamination.py:11` hand-builds a fixture in exactly that shape. That proves the parser handles the shape you assumed; it cannot catch the harness emitting a different one, and nothing else in this repo parses `claude plugin eval --json` to cross-check against. The failure is asymmetric: if the keys are absent you exit 1, which is the safe direction, but if `arms.without` exists while the model's response text lives somewhere other than `graders[].evidence`, the checker prints `ok: N baseline run(s) clean` having inspected grader rationale rather than responses. Worth pinning one real `--json` payload as a fixture, or asserting the expected keys are present rather than defaulting them away.

### Checked and clean

- Merges onto current `main` with no conflicts, and passes the **stricter** validator that landed in #253 — zero errors, 1000 files scanned. Its own CI ran 5 days ago against an older base, so I re-ran this rather than trusting it.
- Registered in all three of `marketplace.json`, `README.md`, `CODEOWNERS`. Version `0.1.0` correct for a new plugin.
- 22 tests pass.
- `format_goal_prompt.py` behaves exactly as `SKILL.md` documents on every path I tried: exit 1 on empty input, exit 1 at `5006 characters; limit is 4000`, and a **warning** with exit 0 on a missing stop clause.
- All three `scaffold.sh` carry `set -euo pipefail` — I flagged this as missing on a first pass and was wrong; it sits on line 4 under a two-line comment.
- The scaffolds generate fixtures in-place specifically so the agent never sees a path into this repo. That is the right instinct and it is the same problem `check_contamination.py` defends.
- No house anti-patterns: no verification scaffolding, no severity pre-filtering, no stale model identifiers, no hardcoded home paths.
- `skills/goal-prompt/agents/openai.yaml` is presentation metadata in the documented location, correctly not treated as an agent definition by the validator.

The contamination problem this PR identifies — that the harness does not sandbox the filesystem, so a no-plugin baseline can read the skill off disk and deflate the delta — generalises well beyond this plugin, and we should lift it into the other eval suites.

dguido:
Pushed `7dbf03a` to this branch with the two fixes from the review above, rather than leaving you to apply them. Both are verified; details so you can check my work rather than take it on trust.

**P1.** `SKILL.md` and `evals/README.md` now invoke scripts with `uv run --no-project` instead of `python3`. I picked that form over a bare `uv run` because the Makefile already uses it in all four of its invocations, though I tested both and both clear the shim. Since the skill runs in the user's repo rather than this one, I also checked `uv run` against a `pyproject.toml` with an unresolvable dependency and against syntactically broken TOML — it survives both. With the shim still on PATH, the exact command `SKILL.md` now documents produces output byte-for-byte identical to what the old `python3` form produced when run shim-free, so this changes reachability and nothing else.

**P2.** Matching is case-folded and `MARKERS` is split into `PATH_MARKERS` and `PHRASE_MARKERS`. A synthetic baseline quoting `**Scope to read first**` as SKILL.md actually writes it is now caught and exits 1, where it previously scored clean.

**Two tests added, both mutation-tested.** Reverting the case-fold fails `test_detects_skill_phrase_in_its_actual_capitalization` and nothing else; renaming the phrase in `SKILL.md` so the marker orphans fails `test_phrase_markers_in_skill` and nothing else. 24 pass, up from 22.

Worth flagging why the original slipped through, because it is a trap rather than an oversight: `test_contaminated_result_exits_nonzero` fed in `"I found scope to read first in SKILL.md"` — the marker's own lowercase spelling instead of the file's. The test asserted against the constant rather than against reality, so 22 green tests sat on a marker that could never fire. The new tests quote `SKILL.md` verbatim and read the file directly, which is why the mutation gate bites.

**Not fixed, and not blocking:** `baseline_texts()` still assumes `cases[].arms.without[].graders[].evidence`, and `_result()` hand-builds a fixture in that same shape, so the tests prove the parser handles your assumption rather than that the harness emits it. Nothing else in this repo parses `claude plugin eval --json` to cross-check. If the keys are absent you exit 1, which is the safe direction, but if `arms.without` exists while the response text lives elsewhere, the checker prints `ok: N baseline run(s) clean` having read grader rationale instead of responses. Pinning one real `--json` payload as a fixture would close it.

The contamination problem this plugin identifies generalises past goal-prompt — the harness not sandboxing the filesystem undermines any with/without delta we measure — and we intend to lift both `check_contamination.py` and the in-place fixture scaffolds into the other eval suites.

Issue #244, opened 2026-08-10
skill-improver: hooks and cancel script depend on `trash`, absent on most Linux distros — loop cannot terminate, cancel fails


lkasvr:
## Environment
- Claude Code 2.1.226, Fedora Linux 44 (`trash`/trash-cli not installed by default)
- Plugin: skill-improver 1.0.1

## Problem
Both `hooks/stop-hook.sh` (4 call sites: lines 66, 81, 113, 123) and `scripts/cancel-skill-improver.sh` (line 61) call `trash "$STATE_FILE"` to remove the session state file. `trash` is not a POSIX/coreutils command — it comes from `trash-cli` (Linux) or Homebrew (macOS) and is not present on stock Fedora/Debian/Ubuntu. `scripts/lib.sh` defines no fallback.

Since both scripts run under `set -euo pipefail`, the missing command aborts the script at exactly the point where the loop should end:

- **Stop hook, completion path** (line 113): even when the `<skill-improvement-complete>` marker is detected, the script dies at `trash` *before* removing the state file and prints no completion JSON. The state file survives, so every subsequent Stop re-enters the loop until `max_iterations` — the loop can never terminate on these systems.
- **Cancel script** (line 61): `/cancel-skill-improver` fails with `trash: command not found` and leaves the state file in place, so the documented escape hatch is also broken. Observed output:
  ```
  .../scripts/cancel-skill-improver.sh: line 61: trash: command not found
  ```

The only way out for the user is manually deleting `.claude/skill-improver.<session>.local.md`.

## Suggested fix
Add a helper in `scripts/lib.sh` and use it everywhere:

```bash
trash() {
  if command -v trash >/dev/null 2>&1; then
    command trash "$@"
  else
    rm -f -- "$@"
  fi
}
```

(The state file is a 158-byte plugin-owned file, so `rm -f` is a safe fallback.)

GrosQuildu:
Fixed in https://github.com/trailofbits/skills/pull/249
Thanks for the report!

Issue #199, opened 2026-07-16
skill-improver: command file shadows the same-named skill — double entry in Claude Code's session skill listing


mikitasazan:
## What happens

The `skill-improver` plugin ships both:

- `commands/skill-improver.md`
- `skills/skill-improver/SKILL.md`

Claude Code (observed on v2.1.211, plugin v1.0.4) surfaces **both** files as separate entries in the model-facing skill listing, under the same display key `skill-improver:skill-improver` — each with its own description pulled from its own frontmatter. So every session pays for two near-duplicate listing entries, and the model sees what looks like a duplicated skill.

## Repro

1. Install `skill-improver@trailofbits` in Claude Code ≥2.1.x.
2. Start a session and inspect the skill listing available to the model.
3. `skill-improver:skill-improver` appears twice — once with the command frontmatter description ("Iteratively reviews and fixes a Claude Code skill until it meets quality standards. Triggers on…"), once with the skill's longer description.

`cancel-skill-improver` is unaffected (command only, no same-named skill).

## Suggestion

Either rename the command (e.g. `improve-skill.md`) or drop one of the two surfaces — since custom commands and skills were merged in Claude Code, a single `skills/skill-improver/` with the desired invocation-control frontmatter should cover both use cases.

Happy to provide more details from my setup if useful.

GrosQuildu:
Fixed in https://github.com/trailofbits/skills/pull/249
Thanks for the report!

PR #249, opened 2026-08-14
skill-improver: remove the trash dependency and the duplicate command surface

GrosQuildu:
Fixes two issues:

- #244: hooks/stop-hook.sh and scripts/cancel-skill-improver.sh called `trash`, which is absent on stock Linux; under `set -e` the stop hook died before removing the state file even when the completion marker was detected, so the loop could never terminate, and the cancel escape hatch failed too. lib.sh now provides remove_state_file(), preferring trash and falling back to rm -f. The helper is not named `trash` because `command -v` resolves shell functions, so a same-named wrapper would always take the trash branch and still die.

- #199: commands/skill-improver.md and skills/skill-improver/SKILL.md both registered as skill-improver:skill-improver, listing twice in the model-facing skill index. Commands are the legacy surface since the commands/skills unification, so the command's path-resolution and setup steps moved into SKILL.md (argument-hint, scoped Bash rule for the setup script) and the command file is gone. Invocation stays /skill-improver. SKILL.md now also tells the model to skip setup on stop-hook continuation prompts so a mid-loop re-trigger cannot start a second parallel session.

Also:
- **`remove_state_file` no longer dies when `trash` exists but fails** (critical): `trash` moved into condition position with `rm -f --` fallback, so a failing trash (network volumes, sandboxed hooks) can't abort the `set -e` callers and leave the state file armed — the #244 loop resurfacing through the trash branch.
- **Codex path for the setup script**: SKILL.md keeps `${CLAUDE_PLUGIN_ROOT}` (officially substituted in skill body and `allowed-tools`) and adds a short resolution fallback (`${CODEX_PLUGIN_ROOT}`, then `find`), same pattern as c-review.
- **Deterministic double-arm guard**: `setup-skill-improver.sh` now refuses to start a second session for a skill that already has an active one; the SKILL.md "skip setup on continuation" prose is now a hint, not the only defense.
- **README causality fix**: natural-language triggering isn't a consequence of shipping as a skill (the command was NL-triggerable too); the skill-vs-command point is now only about listing once in the skill index.


kz-tob (approved):
LGTM

PR #250, opened 2026-08-16
fix(semgrep): three bugs that silently drop entire rulesets

emilushi:
Three bugs in `run-scans.sh` that each cause a third-party ruleset to contribute nothing while the run still reports as complete. Found while auditing a SvelteKit + headless-WordPress repo, where all three fired at once and two required rulesets silently did nothing.

The shared failure mode is what makes these worth fixing: **there is no loud error.** A dropped ruleset shows up as a slightly smaller finding count, which reads as "clean".

### 1. Non-rule YAML in a cloned repo kills the whole scan

`--config` is given the clone root, and semgrep parses every `.yaml`/`.yml` beneath it as a rule. Rule repos ship their own CI config next to their rules, and a workflow's `on: pull_request:` is a null value semgrep rejects - one such file aborts the entire scan with exit 7.

```
$ semgrep --config <clone-of-trailofbits/semgrep-rules> .
semgrep error: null values prohibited
  --> .github/workflows/semgrep-rules-format.yml:3
[ERROR] invalid configuration file found (1 configs were invalid)
$ echo $?
7
```

Same for `elttam/semgrep-rules` via `perf-templates/benchmark-tests.yml`.

Fix: after cloning, prune YAML without a top-level `rules:` key - the invariant that separates a rule file from anything else, so it needs no per-repo knowledge. It also drops `*.test.yaml` fixtures, which are rule test inputs. Measured on this prune alone: keeps 118/145 files for trailofbits and 80/94 for elttam, losing no real rule.

### 2. `mode: join` rules crash the scanner

Only visible once fix 1 lets elttam load its rules. `rules/generic/jsp-likely-xss.yaml` crashes semgrep 1.173:

```
File ".../semgrep/join_rule.py", line 153, in evaluate_condition
    return getattr(A, property_a) == getattr(B, property_b)
AttributeError: type object 'JspScriptletParam' has no attribute '$VAR'
```

That's a hard process failure, not a rule-level error — the batch dies and writes no output, losing every other rule in the same invocation. `join` is experimental and rare, so pruning costs far less than losing the run. It takes elttam to 79 of its 94 YAML files.

A join rule is identified by two keys, not one: `mode: join` **and** the `join:` block that mode requires, each matched as a YAML key on its own line over `*.yaml`/`*.yml` only. Matching the bare string anywhere in any file would delete a rule whose `message:` quotes the docs, and with it every sibling rule in the same file. A rule that names the mode without the block is not a valid join rule; semgrep rejects it per-rule, which fix 3 now survives, so it is left in place.

### 3. Exit 2 treated as "no scan happened"

The gate is `rc -eq 0 || rc -eq 1`, commented as "7 means the config would not load and 2 a bad argument; in both cases no scan happened". Exit 2 is also what semgrep returns when *individual rules* fail to compile while the run completes and writes full JSON and SARIF.

Two rulesets were discarded by this:

- `elttam/semgrep-rules` — 12 Java rules current semgrep can't parse; **107 others ran fine** over ts/php/js/yaml.
- `apiiro/malicious-code-ruleset` — filed as failed while its own captured log read `✅ Scan completed successfully • Findings: 51`. Those findings never reached the merged SARIF, and nothing in `scans.json` indicated a loss.

Fix: allow exit 2 through and decide on the artifacts, which distinguish the two things exit 2 means. A config that will not load writes nothing, so the existing `-s` checks reject it unaided; a run whose rules partly failed writes complete output and is kept, flagged `partial`. Anything outside 0/1/2 stays fatal regardless of what was written.

**Worth knowing, and measured rather than assumed: the exit code for an unloadable config depends on the output flags.** Same rules directory and target, semgrep 1.173, back to back:

```
semgrep --config rules target                               -> 7, nothing written
semgrep --config rules -o out.json --sarif-output=out.sarif -> 2, nothing written
```

This script uses the second form, so the repro in section 1 above surfaces here as exit 2, not 7. That is also why the fatal branch is belt-and-braces rather than the load-bearing part — but it costs nothing, and it keeps a future semgrep that writes an empty result set alongside a hard failure from reading as a clean scan.

Covered by tests either way: exit 2 with complete output is kept, counted and flagged `partial`; exit 2 with no output still fails; exit 7 stays fatal.

### 4. `partial` is now read, not just written

A ruleset with 12 uncompilable rules that reports as an unqualified success is the same class of bug as one that is discarded — better, but still a coverage loss nothing surfaces. So `partial` is plumbed through every consumer:

- `workflows/semgrep-scan.js` — a required `partial` field on `SCAN_SCHEMA`, described as a subset of `succeeded` rather than an addition to it; the scan prompt reads it with `jq '[.scans[] | select(.partial)] | length'`; the run log reports it; the report step gets a "Ran Partially" instruction; the workflow returns it.
- `skills/semgrep/workflows/scan-workflow.md` — a `scans[].partial` row in the field table marked **Must be shown**, `exitCode` documented alongside it, and a `### Ran Partially:` section in the report template.
- `skills/semgrep/SKILL.md` — a matching line in the Success Criteria checklist.

### Neither prune can fail quietly

Both prunes delete files, so both are checked rather than suppressed. `2>/dev/null || true` is gone: a prune that cannot run drops the ruleset into `.skipped` with a reason and deletes the half-pruned clone rather than scanning it, and every clone reports what it took —

```
elttam-semgrep-rules: pruned 14 non-rule YAML file(s) and 1 join-mode rule file(s)
```

A run whose filter stopped matching prints `0` there, which is the signal a silent no-op never gave.

### Verification

Same three rulesets, before and after, via `run-scans.sh` itself:

| ruleset | before | after |
| --- | --- | --- |
| `trailofbits/semgrep-rules` | exit 7, discarded | **11 findings**, 205 files, `partial: false`, exit 0 |
| `elttam/semgrep-rules` | discarded | 0 findings, 205 files, `partial: true`, exit 2 |
| `apiiro/malicious-code-ruleset` | discarded (findings lost) | **54 findings**, 109 files, `partial: true`, exit 2 |

`3 scan(s) succeeded, 0 failed, 0 skipped` - previously all three failed.

Both prunes now have the fixture AGENTS.md asks for. A new `mixed` git-stub mode builds the tree a real rule repo ships — a rule file, the repo's own CI workflow, a `.test.yaml` fixture, a join rule, a decoy rule whose `message` quotes `mode: join` next to a sibling rule, and a README that mentions it — and asserts which files survive, that the ruleset still scans, and that the prune counts are reported. Reverting to the previous broad `grep -rlE` turns five of those assertions red.

`shellcheck` and `shfmt -i 2 -ci` clean. `run_scan_tests.sh` at 93 assertions (up from 78 on main), `run_workflow_tests.sh` at 64 plus its 65-assertion self-test, with two new mutations — "stop reporting partial rulesets" and "stop reading the partial count" — both caught. `make lint shell js-tests validate` green and `validate_plugin_metadata.py --base-ref main` exits 0.

Rebased onto current `main`, and `static-analysis` bumped to 1.4.3 in **both** `plugin.json` and `marketplace.json` - main reached 1.4.2 after the 1.4.1 bump on this branch.

### Notes

- The prunes delete files inside the tool's own `repos/` clone directory, which is already `rm -rf`'d and re-cloned per run, so nothing user-owned is touched.
- Fix 3 changes the shape of `scans.json` entries (`partial` and `exitCode` added). If any downstream consumer asserts an exact key set, that's the thing to check.
- The join prune deletes whole files, so a file holding a join rule loses its sibling rules too. Removing one rule from a multi-rule YAML needs a parser this script does not have, and losing a file beats losing the batch.
- I left the `third_party` https-URL validation alone, though allowing local rule directories would also let callers pre-clean their own rules. Happy to add that separately if it's wanted.


kumarak:
@emilushi, could you fix the CI errors?

kz-tob:
static-analysis: version 1.4.0 is not greater than 1.4.0 at the merge base; Bump it, clients only pull an update when the number increases. If another PR bumped this plugin after you branched, rebase first

kz-tob (changes requested):
I pushed the version bump — main had moved to 1.4.0 since you opened this, so it's 1.4.1 in both plugin.json and marketplace.json now.

One gap to fix: nothing reads the new partial key. workflows/semgrep-scan.js still counts .scans | length as succeeded, and the scans.json field table in skills/semgrep/workflows/scan-workflow.md documents only json/sarif/findings/filesScanned. So a ruleset with 12 uncompilable rules now reports as an unqualified success — better than being discarded, but that doc's own line applies:
  "a run that covered four of nine rulesets reads exactly like one that covered four of four unless you say otherwise." Could you add a partial row to the table and a line in the report step telling it to surface them?

  Fixes 1 and 2 need a bit more before I'd merge them:

  1. No tests on either, and they're the two that delete files. find … ! -exec grep -q '^rules:' {} \; -delete 2>/dev/null || true carries both stderr suppression and || true, so if it ever breaks it silently no-ops and the exit-7 bug returns with nothing to show for it. AGENTS.md asks for a fixture proving a filter still detects its target — a small tempdir with one rule file, one workflow YAML, one .test.yaml would do it.
  2. The join prune is too broad. grep -rlE 'mode:[[:space:]]*join' "$dest" greps every file rather than just YAML, and matches the string anywhere rather than as a rule key — a rule whose pattern: or message: contains mode: join gets deleted, and since one file can hold many rules it takes the siblings with it. Restricting to *.yaml/*.yml and matching the key at rule level would keep it targeted.

emilushi:
@kz-tob both points addressed, and thanks for pushing the 1.4.1 bump. Rebased on current `main` (now 1.4.2), so this carries **1.4.3** in `plugin.json` and `marketplace.json`.

**`partial` is now read, not just written.** `workflows/semgrep-scan.js` gets a required `partial` field on `SCAN_SCHEMA`, described as a subset of `succeeded` rather than an addition to it, and the scan prompt tells the agent to read it (`jq '[.scans[] | select(.partial)] | length'`). It couldn't report a field it was never told to look at. The run log reports it, the report step gets a "Ran Partially" instruction, and the workflow returns it. In `scan-workflow.md` the field table gains a `scans[].partial` row marked **Must be shown** (with `exitCode` documented alongside it, since I was adding both keys), the "four of nine rulesets" paragraph now covers partials, and the report template gains a `### Ran Partially:` section. `SKILL.md`'s Success Criteria checklist got a matching line.

Pinned in `workflow-harness.js`: assertions that the count is carried through, that the scan prompt asks for it, and that the report prompt surfaces it, plus two mutations. "stop reporting partial rulesets" and "stop reading the partial count", both caught by the self-test.

**Both prunes have a fixture now.** A new `mixed` git-stub mode builds the tree you described plus the case that matters most - a rule file, the repo's own CI workflow, a `.test.yaml` fixture, a real join rule, a decoy rule whose `message` quotes `mode: join` with a sibling rule beside it, and a README that mentions it. Nine assertions on what survives, that the ruleset still scans, and on the reported prune counts. I mutation-tested them: reverting to the old broad `grep -rlE` turns five red. Suite is at 93, up from 78 on main.

**The silent no-op is gone.** `2>/dev/null || true` is removed; both prunes go through a `prune_yaml` helper whose status is checked. A prune that cannot run drops the ruleset into `.skipped` with a reason and deletes the half-pruned clone rather than scanning it. And every clone now prints what it took

```
elttam-semgrep-rules: pruned 14 non-rule YAML file(s) and 1 join-mode rule file(s)
```

- reported even as zeroes, so a filter that stopped matching no longer looks identical to a repo with nothing to prune.

**The join prune is narrowed.** It now requires *two* keys: `mode: join` and the `join:` block that mode needs - each anchored as a YAML key on its own line (optionally the first key of a list item, optionally quoted, trailing comment allowed), over `*.yaml`/`*.yml` only. Verified against the real repos: still hits `elttam/rules/generic/jsp-likely-xss.yaml`, matches nothing in `trailofbits/semgrep-rules`. A rule that names the mode without the block is not a valid join rule - semgrep rejects it per-rule, which the exit-2 handling now survives - so it is left alone. I also dropped `xargs -r`, which BSD xargs doesn't reliably take.

Two things worth flagging:

- The prune still deletes whole files, so a file holding a join rule loses its siblings. Removing one rule from a multi-rule YAML needs a parser this script doesn't have, and losing one file beats losing the batch - but the anchoring means it now only happens to files that really do carry a join rule.
- One correction to my original numbers: 80/94 for elttam is the rules-key prune alone; the join prune then takes its one join rule, leaving **79**. The comment in the script and the PR body both say so now.

`shellcheck` and `shfmt -i 2 -ci` clean, `make lint shell js-tests validate` green, `validate_plugin_metadata.py --base-ref main` exits 0.

kz-tob (approved):

PR #253, opened 2026-08-18
Back three AGENTS.md enforcement claims with validator checks

dguido:
AGENTS.md lists three checks under **"What the validator enforces, so you do not have to"** — the heading that tells contributors and Claude to skip checking them by hand. The validator performed none of them. Found while taking a mechanical inventory of the repo ahead of a wider quality pass, by probing the validator with single-defect fixtures rather than reading it.

## What changes

**Hardcoded `/Users/…` and `/home/…` paths** move out of `.github/workflows/validate.yml` and into `find_hardcoded_paths()`. The scan was CI-only, so `make check` and pre-commit never ran it. It also used `grep -rPn`; PCRE is a GNU extension, so that command could not run on macOS even by hand. Python's `re` supports the negative lookbehind natively, so the pattern ports unchanged. Same file types, same `*-shim.bats` exemption, same `/path/to` and `/home/vscode` placeholders. The CI step's anti-vacuity guard comes across too: zero files scanned is a hard failure, mirroring the existing `refs_checked == 0` guard.

**Command files** are now checked for the `tools:`/`allowed-tools:` inversion. `validate_agent_frontmatter()` opened agent files and SKILL.md files only, so the "and commands" half of that claim was unbacked and nothing anywhere checked it. Renamed to `validate_tools_frontmatter()` now that it covers all three file types.

**`subagent_type`** resolves against a repo-wide agent registry instead of one plugin's own agents. The old check returned early when a plugin had no `agents/` directory, and otherwise only flagged a bare name matching that same plugin's agent — so a bare name borrowed from another plugin, or naming an agent that does not exist, passed clean. Both now report, and the cross-plugin case names the right namespace in the message.

AGENTS.md also gains three checks the validator already enforced but the list never mentioned: `plugin.json` `name` matching the directory name, marketplace `source`/`description` parity, and dependabot lockfiles.

## Evidence

All three flag **zero violations against the repo as it stands**, so nothing here changes the build's verdict today. Each was verified by planting the defect and confirming rejection:

- `plugins/gh-cli/probe-path.sh:2 hardcodes an absolute user path: cd /Users/Someone/cc/skills`
- `skill-improver: commands/probe.md uses 'tools:'; commands must use 'allowed-tools:'`
- `gh-cli: probe-agent.md: subagent_type 'poc-builder' is not namespaced; use 'fp-check:poc-builder'`
- `gh-cli: probe-agent.md: subagent_type 'nonexistent-thing' names no agent in this repo and is not a builtin`

Negative controls pass silently: a command using `allowed-tools:`, a `/home/vscode` container path, and a `*-shim.bats` fixture holding a literal `/home/user` path.

Self-test goes from 34 assertions to 43. `SELF_TEST_MINIMUM` is now the exact count rather than a floor 14 below it, so losing a fixture has to be a deliberate edit.

Full scan after the change: `no errors (395 references resolved, 980 files scanned for hardcoded paths)`, with the same 8 skill-length warnings as before.

### DIFF
--- a/.github/scripts/validate_plugin_metadata.py
+++ b/.github/scripts/validate_plugin_metadata.py
@@ -84,18 +84,28 @@ SEMVER_PATTERN = re.compile(r"^(\d+)\.(\d+)\.(\d+)")
 # letters does not match. Python's re supports lookbehind natively, which is the whole
 # reason this lives here rather than in the CI grep it replaces: `grep -P` is a GNU
 # extension and this repo's house rule is POSIX ERE, portable to BSD grep on macOS.
-HARDCODED_PATH_PATTERN = re.compile(r"(?<![a-zA-Z])(?:/home/[a-z]|/Users/[A-Z])")
+#
+# Both branches accept either case. The version inherited from the CI step matched
+# `/Users/[A-Z]` only, so every conventionally-lowercase macOS account name — the
+# overwhelmingly common form, `/Users/alice`, and this repo's own `/Users/user` — went
+# undetected while the `/home/` branch caught its equivalents. The check missed the
+# thing it exists to find.
+HARDCODED_PATH_PATTERN = re.compile(r"(?<![a-zA-Z])(?:/home/[A-Za-z]|/Users/[A-Za-z])")
 # Shell, bats, yaml and toml included deliberately: test fixtures and install scripts
 # are exactly where absolute paths hide.
 HARDCODED_PATH_SUFFIXES = (".md", ".py", ".json", ".sh", ".bats", ".yml", ".toml")
 # The shim suites need literal /home/user paths to test what they exist to test.
 HARDCODED_PATH_EXEMPT_SUFFIX = "-shim.bats"
-# Placeholders and container-image paths, not any individual's home directory.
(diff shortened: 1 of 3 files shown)


dguido:
Fixed in fea2d5a. This one was worse than the `nit` label suggests, so recording the detail.

The gap was not only that `/Users/Alice` matched and `/Users/alice` did not. The pattern missed **this repository's own path** — `/Users/user/cc/skills` — along with every conventionally-lowercase macOS account name, which is essentially all of them. A check whose stated job is finding leaked absolute home paths could not see the most likely one.

My mutation test passed only because I happened to plant `/Users/Someone` with a capital S, the single spelling the pattern could see. That is a test agreeing with the bug rather than catching it, which is exactly the failure class this repo's `--self-test` discipline exists to prevent.

Both branches now accept either case. Widening surfaced exactly one new match across the repo, and it is a false positive: `plugins/c-review/skills/c-review/SKILL.md:120` uses `"/Users/me/My Repo"` to illustrate that a path containing a space must stay quoted. That and `/Users/Shared`, a genuine macOS system directory, are now in the placeholder list. `/Users/user` deliberately is **not** — it is a real account name here, not a stand-in.

Self-test goes 43 → 45: one fixture asserting the lowercase path is caught, one asserting `/Users/Shared` is not.

PR #255, opened 2026-08-18
Narrow the modern-python shims to the commands uv run replaces

### LINKED ISSUE #207: modern-python shims break prek hook installation and `python3 -`
The `modern-python` plugin's shims intercept two invocations that are not package-management operations, and both break real tooling. Found while adding `make check` in #206.

## 1. `uv pip install` is rejected, which stops prek installing hook environments

```
error: Failed to install hook `check-json`
  caused by: Command `~/.claude/plugins/cache/trailofbits/modern-python/1.5.2/hooks/shims/uv pip install --project / --directory ~/.cache/prek/repos/... .` exited with an error:
ERROR: `uv pip` is the legacy interface. Use instead:
uv add <package>       # instead of uv pip install
```

The advice is right for a human managing project dependencies and wrong for prek, which is building an isolated hook environment and has no project to `uv add` into. Effect: `git commit` fails outright in any repo using prek with hooks that need a Python env. I had to use `--no-verify` for every commit in #206.

## 2. `python3 -` (script on stdin) is redirected to `uv run`

```
ERROR: Use `uv run python3 - <args>` instead of `python3 - <args>`
```

`python3 -` reads a program from stdin. There is no file, no dependency resolution to do, and `uv run python3 -` is not a drop-in replacement in a pipeline. Effect: `plugins/zeroize-audit/tests/rust-regression/run_smoke.sh` fails at line 72 on any machine with this plugin installed. That is why `make shell-suites` is deliberately not part of `make check` in #206 — the suite is fine, the shim is not.

## Suggested fix

Narrow both shims to the cases the guidance is actually about:

- Let `uv pip` through when it carries `--project`, `--directory`, or `--target` — those signal tooling managing an environment it owns, not a human installing into a project.
- Do not intercept `python3 -`, `python3 -c`, or `python3 -m` at all. None of them resolve dependencies from a script path, which is the thing `uv run` exists to handle.

More generally, the pattern to avoid here is a hook that blocks a command it cannot distinguish from the one it means to catch. Both of these fire on a substring rather than on intent.

cc @dguido

dguido:
Closes #207.

The shims sit on PATH, so they intercept every subprocess *any tool* spawns, not only what Claude types. Two of the intercepted invocations are not package management, and blocking them broke real tooling.

## `uv pip` with a tool-managed environment

Passes through when the invocation carries `--project`, `--directory` or `--target`. Those say a tool is building an environment it owns, where `uv add` is not advice it can act on — prek installs every hook with `uv pip install --project / --directory <cache>` and has no project to add to. The refusal made `git commit` fail outright in any repo whose hooks need a Python environment. A bare `uv pip install requests` is still refused.

## `python -c`, `python -m <module>`, `python -`

These reach the real interpreter now. None resolves a script against a project's dependencies, which is the thing `uv run` exists to do, and `uv run python3 -` is not a drop-in replacement inside a pipeline. Still intercepted: `python -m pip` (that *is* the foot-gun), bare `python`, and `python script.py`.

Passing anything through is new for the python shim, which previously ended every branch in `exit 1`, so it gains the same skip-my-own-directory PATH walk the uv shim already had. That walk uses parameter expansion instead of `basename`, because the one case where it must report failure is a PATH holding nothing but the shim — where shelling out to coreutils fails first, with a confusing error about `basename` and the right exit status for the wrong reason.

## Evidence

Each symptom in #207 run against the old shim and the new one:

| | old shim | new shim |
|---|---|---|
| zeroize-audit rust-regression smoke test | `run_smoke.sh: FAILED at line 72` | `Rust regression smoke checks passed.` |
| prek hook install, cold cache | `Failed to install hook` | `check json ... Passed` |

Nine behaviors spot-checked by hand, all correct: `uv --version` and `uv pip list --directory /tmp` pass through, `uv pip install requests` refuses, `python -c 'print(1+1)'` prints 2, `python -m json.tool --help` works, `echo 'print(3+3)' | python3 -` prints 6, and `python -m pip install x`, bare `python` and `python script.py` all refuse.

bats goes from 19 cases to 38. Five python cases were **inverted rather than deleted** — the ones asserting `-c` and `-m` are refused now assert they run — so the suite still describes the contract rather than having holes cut in it. Both suites declare `bats_require_minimum_version 1.5.0`, which also clears the BW01 warning on the exit-127 case by declaring the expected status with `run -127`.

## What this does not fix

`make shell-suites` still fails on a machine with this plugin installed, but no longer for the reason #207 gives. The remaining failure is `plugins/variant-analysis/tests/` invoking `python3 <script>.py`, which the shim intercepts **by design** — a bare script run is exactly what `uv run python` replaces. With no shim on PATH the whole target passes. That one belongs to variant-analysis, so AGENTS.md's note is corrected rather than deleted.

Note for reviewers: the shims that ship here take effect only after the plugin updates to 1.6.0, so a machine still on 1.5.3 keeps the old behavior.

Version bumped 1.5.3 → 1.6.0, minor because this changes behavior for everyone who has the plugin installed.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -241,7 +241,7 @@
     },
     {
       "name": "modern-python",
-      "version": "1.5.3",
+      "version": "1.6.0",
       "description": "Modern Python best practices. Use when creating new Python projects, and writing Python scripts, or migrating existing projects from legacy tools.",
       "author": {
         "name": "William Tan",
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -239,9 +239,13 @@ is strong evidence and not a guarantee:
   `pre-commit run -a`) to cover those locally.
 - **the version-increment check**, which needs a base ref to diff against and so has
   no meaning outside a PR.
-- **`make shell-suites`**, which is a target but not part of `check`: it fails on any
-  machine with the `modern-python` plugin installed, because its shim intercepts the
-  `python3 -` that zeroize-audit's suite uses (#207).
+- **`make shell-suites`**, which is a target but not part of `check`: it still fails on a
+  machine with the `modern-python` plugin installed, though no longer for the reason
+  #207 describes. The `python3 -` interception that broke zeroize-audit is gone as of
+  modern-python 1.6.0. What remains is `plugins/variant-analysis/tests/` invoking
+  `python3 <script>.py`, which the shim intercepts *by design* — a bare script run is
+  exactly what `uv run python` replaces. That one is variant-analysis's to fix. With no
+  shim on PATH the whole target passes.
 
 Both scan every plugin; the validator is not scoped down in CI. Only the
 version-increment check is limited to the plugins a branch touched, and it is the one
--- a/plugins/modern-python/.claude-plugin/plugin.json
+++ b/plugins/modern-python/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "modern-python",
-  "version": "1.5.3",
+  "version": "1.6.0",
   "description": "Modern Python best practices. Use when creating new Python projects, and writing Python scripts, or migrating existing projects from legacy tools.",
   "author": {
     "name": "William Tan",
--- a/plugins/modern-python/README.md
+++ b/plugins/modern-python/README.md
@@ -39,15 +39,17 @@ Modern Python tooling and best practices using uv, ruff, ty, and pytest. Based o
 
 This plugin includes a `SessionStart` hook that prepends PATH shims for `python`, `pip`, `pipx`, and `uv`. When Claude runs a bare `python`, `pip`, or `pipx` command, the shell resolves to the shim, which prints an error with the correct `uv` alternative and exits non-zero. The suggested alternative always uses the exact command name `python` (never `python3`) so it also works outside a project; see the header comment in [`hooks/shims/python`](hooks/shims/python) for the full rationale.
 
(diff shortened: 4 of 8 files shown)


dguido:
Both findings addressed in 3cdc048, one of them fully and one deliberately in part.

**`-t` short flag** — fixed. `uv pip install --help` does document `-t, --target <TARGET>`, so the same tool-managed install was allowed or refused depending on spelling. Exempted alongside the long form, with a bats case.

**`$1`-only matching, python half** — fixed, and this was the more useful catch. `python -u -c 'code'` was refused while `python -c 'code'` ran, which made the decision depend on argument order rather than on what the command actually does. The shim now steps over interpreter flags to find the mode selector, giving `-W`, `-X` and `--check-hash-based-pycs` the two argument slots they consume. `-u -m pip` is still refused, and `-u script.py` still is too, since a script path is what `uv run` replaces no matter what precedes it.

**`$1`-only matching, uv half** — not fixed, on purpose. `uv --no-progress pip install requests` does slip past the refusal. Doing this correctly requires knowing which uv global flags take a separate value: a naive "first token not starting with `-`" reads `always` as the subcommand in `uv --color always pip install`, and scanning every argument for a bare `pip` would refuse `uv add pip`, which is a legitimate command. Both would be new bugs in a PATH shim that gates every invocation. The current failure mode is a missed nudge, not a breakage — the real `uv` runs and behaves correctly — and this PR's purpose is to refuse *less*, so widening what it refuses does not belong here.

bats 38 → 43. Both #207 regressions re-verified after the restructure, since it touched the code path that makes them work: zeroize-audit's smoke test passes, and prek installs hooks from a cold cache.

PR #256, opened 2026-08-19
Collapse the review into one xhigh opus tier, and stop invented evidence

dguido:
Audit of what was actually reviewing #253-#255, plus the cleanup that fell out of it.

## The model was never specified

`claude_review.sh` passed `--effort` but no `--model`, so the reviewer was whatever the CLI resolved to at run time, and nothing in the logs recorded which. The CLI version was pinned specifically so CI could not change without a commit, while the larger lever on review quality was left floating.

Now pinned to `opus`, with the run line recording all three:

```
Reviewing trailofbits/skills#256 (model=opus effort=xhigh cli=2.1.235 (Claude Code))
```

## The review claimed to run things it cannot run

The allowlist is `gh pr` reads, `git log`, `git diff`, `Read`, `Grep`, `Glob` — no interpreter, no test runner, no shell. Across the three PRs it reviewed:

- **#253**: "I verified this directly:" and a Python transcript. Its regex, `re.compile(r"(?<[a-zA-Z])...")`, raises `re.error: unknown extension ?<[` — it cannot compile, so that session never ran.
- **#254**: "I ran the full test_workflow_contract.py suite locally (node v22.22.0) and confirmed all 49 tests pass." The suite reports 82 passed, 2 skipped.
- **#255**: "verified directly by three independent verifiers running the shim."

Every conclusion was correct. Every proof was invented. That is worse than showing no proof, because a reader cannot tell which findings were actually checked.

The cause was a prompt gap rather than a reasoning failure: the **deep** prompt already said "You cannot execute anything," and the tier that actually ran never did. The prompt now states the tool reality and forbids reporting output no tool produced, while explicitly still permitting reasoning about what an input would do.

## One tier instead of two

The split was a cheap pass on every push plus an xhigh adversarial pass behind a `deep-review` label. It did not survive contact: **the label was never created**, so the deep tier skipped 9 times and ran zero, and every review this repo has received came from the cheap tier.

Effort goes `low` → `xhigh`, and with both tiers then identical apart from scope there was no reason to gate the better scope behind a label somebody has to remember to apply. So there is one job, taking the deep tier's prompt and tools (reads beyond the diff, git history, checks the PR's claims against the files) and the cheap tier's `--edit-last` posting, which matters now that it runs on every push.

`fast` is gone from the check name, the script's tier argument, and the concurrency key — it was already inaccurate once effort went to xhigh. I confirmed nothing depended on it: the ruleset on `main` requires `license/cla`, `Validate plugins and skills`, `Pre-commit` and `Shell (bats)`, not this check. The label-keyed concurrency group goes too, since it existed only to stop a push cancelling an in-flight deep review.

## The CLI is now unpinned, on purpose

Both here and in the loadability check in `validate.yml`. For loadability that is arguably the point: the version worth proving plugins load against is the one users are running, not one frozen at the last manual bump. The trade is that a release can change behavior with no commit here, which is why the resolved version is now logged. zizmor still passes; the existing `adhoc-packages` ignores cover it.

## Also

The prompt now says to rank on consequence rather than diff size. The `/Users/[A-Z]` gap in #253 was filed as a nit, but it made a brand-new checker miss the most likely path it exists to catch, including this repository's own.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,7 +21,9 @@
 #
 # Python deps declared with PEP 723 inline metadata (the common pattern in this
 # repo, including .github/scripts/) have no Dependabot ecosystem at all and stay
-# manual. So do the pinned npm CLI versions in .github/workflows/validate.yml.
+# manual. So does the pinned Codex CLI in .github/workflows/validate.yml — the
+# Claude Code CLI beside it is deliberately unpinned instead, on the grounds that a
+# pin nobody bumps is worse than a break that announces itself.
 version: 2
 
 updates:
--- a/.github/scripts/claude_review.sh
+++ b/.github/scripts/claude_review.sh
@@ -26,9 +26,24 @@ TIER="${1:?usage: claude_review.sh <fast|deep>}"
 : "${PR_NUMBER:?PR_NUMBER is required}"
 : "${REPO:?REPO is required}"
 
+# The model is pinned explicitly. Left unset, `claude` resolves whatever the CLI
+# defaults to at run time, so the reviewer could change underneath this repository
+# with no commit and nothing in the logs would say which model had reviewed the diff.
+#
+# Both tiers now run opus at xhigh: every pull request gets the strongest review, not
+# only the ones somebody remembers to flag. What still separates them is scope — the
+# deep tier reads beyond the diff, gets git history and a longer budget, and its
+# prompt is adversarial. The `fast` name is left alone because it is the check name
+# branch protection matches on; it describes the trigger, not the effort.
 case "$TIER" in
-  fast) EFFORT="low" ;;
-  deep) EFFORT="xhigh" ;;
+  fast)
+    EFFORT="xhigh"
+    MODEL="opus"
+    ;;
+  deep)
+    EFFORT="xhigh"
+    MODEL="opus"
+    ;;
   *)
     echo "unknown tier: $TIER (expected fast or deep)" >&2
     exit 2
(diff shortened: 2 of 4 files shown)


dguido:
Acted on the review in e3aa66f. Four findings fixed, one reverted on your argument, three acknowledged without change.

**P2, API key in the unpinned install step** — fixed, and you were right that this is the sharpest one. Presence is now reduced to a boolean in its own step, so the install step carries no secret at all. That is what makes `@latest` defensible there rather than merely convenient.

**P3, the posting guard counting comments it did not write** — fixed and measured. On this PR's siblings the old query counts 2 and the bot-only query counts 1, because I had commented on both by hand. So the failure you described was live, not theoretical: a maintainer reply could have stood in for a review that never posted. Also paginated.

**P3, `AGENTS.md` reaching the reviewer** — fixed, and this was the best catch in the review. `CLAUDE.md` is literally `@AGENTS.md`, so it loads as project instructions telling the model to run `make check`, run `prek run -a`, and consult a `claude-code-guide` subagent, none of which it can do. That is the fabrication channel this PR exists to close, arriving by a route the prompt never addressed. The prompt now names those files and says they are addressed to someone with a full toolchain.

**P3, unpinning a required check** — reverted. `validate.yml` goes back to a pinned CLI. Unpinning it was my extension rather than what was asked for, and your placement argument is correct: that job is a required check, so a release renaming a field in `plugin list --json` reddens every open PR at once with no commit to explain it. The review job can go red harmlessly. This also resolved two of the P4s for free — Codex and Claude are pinned together again, and dependabot.yml's note about "the pinned npm CLI versions" is accurate again.

**P3, the `opus` alias** — comment corrected rather than the value. The alias tracking new Opus releases is intended, the same trade as the unpinned CLI; the comment claimed an attributability it does not deliver, so it now says the log line is what narrows a surprising review to a date.

**P4, "only `gh pr` reads"** — fixed. Inaccurate on its face: `gh pr comment` is a write and is mandated fourteen lines later, and `git log`/`git diff` are execution.

**P4, the ruleset** — checked, and your caution was warranted but the claim holds. `gh api repos/trailofbits/skills/rulesets` gives required checks of `license/cla`, `Validate plugins and skills`, `Pre-commit` and `Shell (bats)`. The review check is not among them, so the rename is safe.

**P3, "the prompt now running on every push has never run"** — correct and unfixable in sequence; this run is its first. Noting it rather than pretending otherwise. On the related flag about "finish with an explicit list of what you checked" sitting near AGENTS.md's warning on verification scaffolding: I read it the same way you did, as a reporting requirement, and the "Checked and clean" section it produced here is the most useful part of the review. Keeping it, and it is the first thing to cut if quality regresses.

Left alone: `--edit-last` targeting the newest bot comment rather than the newest review (latent while this is the only workflow with `pull-requests: write`), the `cli=)` log on a missing binary (the next line fails loudly anyway), and the identical-body PATCH question, which I also could not resolve.

One correction to the review. On #254 you wrote that "test_workflow_contract.py is pytest, not node, so (node v22.22.0) was incoherent regardless of the count." The file is pytest and it shells out to `node --test` for two of its suites, so naming a node version there was reasonable; the incoherence was the claim to have run it at all, plus a count of 49 against an actual 82 passed and 2 skipped.

dguido:
Reopening the pinning question on `validate.yml`, deliberately, so the next review does not read it as a regression.

The P3 argument for pinning the loadability check was right about the mechanics: it is a required check, so a Claude Code release that breaks plugin loading reddens every open PR at once with no commit here to explain it. I reverted to a pin on that basis. That call has now been reversed on the owner's judgment, and the reasoning is worth recording.

A pin nobody bumps does not stay accurate, it drifts — until CI is proving plugins load against a version no user is running, which is the single thing that check exists to establish. That failure is silent and it passes. The unpinned failure is loud and it blocks. Between a check that quietly stops meaning anything and one that stops the line when its subject actually breaks, the loud one is preferable. Dependabot does not track npm CLIs installed this way, so the choice was only ever between a live version and a stale one, never a maintained one.

`@anthropic-ai/claude-code@latest` in both workflows now. The Codex CLI beside it stays pinned at `0.146.0`; it is the same class of manual pin and probably deserves the same treatment, but unpinning another vendor's CLI was not part of this change. `dependabot.yml`'s note is corrected to singular and says which one is pinned and why the other is not.

The trade is accepted rather than overlooked: if a Claude Code release breaks `plugin list --json`, this repo's merges stop until someone looks. That is the intended behaviour.

PR #258, opened 2026-08-19
Make every documented command runnable under our own python shims

dguido:
Our `modern-python` plugin ships PATH shims that refuse `python <script>`, `pip install`, `python -m pip` and `uv pip install`. **Twelve other plugins in this marketplace issued exactly those forms**, so installing our own plugin broke our own skills — with CI green the whole time.

The worst case is not theoretical. `c-review` and `rust-review` both call their Phase 4 planner as `python3 "${PLUGIN_ROOT}/scripts/build_run_plan.py"`, so with the shim installed **every run died before spawning a worker**. Verified both directions with the shim on PATH:

| | exit |
|---|---|
| `uv run --no-project .../build_run_plan.py --help` | 0 |
| `python3 .../build_run_plan.py --help` | 1 (refused) |

## Why this is not one substitution

Phase 1's reading pass named 16 skills. A mechanical sweep found **96 candidate lines across 44 files**, and adding shell scripts to the sweep found **10 more the docs-only pass had missed** — which is the argument for the check at the end rather than a hand-fix.

Four classes needed different treatment, and in two of them a careless rewrite would have been worse than the breakage:

- **Our own scripts** → `uv run --no-project <script>`. Not bare `uv run`: these execute inside the *target* repo, which may be a Python project that cannot resolve. Verified against a deliberately broken `pyproject.toml`, and against `validate_artifacts.py`'s sibling import of `generate_sarif`.
- **Package installs** → `uv add` for a dependency, `uv tool install` for a CLI, `uv sync` for a project's own editable install.
- **Third-party CLIs we merely document** (OSS-Fuzz's `infra/helper.py`, yarGen) → `uv run --no-project python <script>`, preserving upstream's exact semantics rather than handing their script an environment we manage. Verbose in a command table; correct everywhere else.
- **atheris's instrumented build** keeps its source build as `uv add --no-binary-package cbor2`. Dropping that flag would silently produce an *uninstrumented* fuzzer, which is worse than a visible failure. Its prose said "the `--no-binary` flag" and now names the flag it actually uses.

## Two factual corrections that fell out

`pip install caracal` was wrong twice over. Caracal has `Cargo.toml` and `Cargo.lock` at its root, so it is a Rust tool — a `uv tool install` would have fetched an unrelated PyPI package. It is now upstream's own `cargo install --git https://github.com/crytic/caracal --profile release --force`, quoted from their README.

`pip install uv` cannot bootstrap uv under a shim that intercepts pip, so `culture-index` now points at the official installer.

## Thirteen lines stay, each deliberately

Dockerfile `RUN` lines and oss-fuzz's `build.sh` run **inside containers**, where our shims are absent. `codeql`'s pip calls install the *analysed* project's dependencies, and that project is arbitrary and may not be uv-managed — substituting uv there would change its build and risk an incomplete database. `trailmark`'s dispatch skills must keep saying "Do NOT run `pip install`". And `modern-python` documents what it intercepts.

Each exemption is now **explicit** rather than an accident of regex: structural for `dockerfile` fences, and an `allow-legacy-python: <reason>` comment that scopes to its code block for the rest.

## Side effect: `make shell-suites` passes again

`plugins/variant-analysis/tests/` no longer invokes `python3 <script>.py`. Exit 0 with the 1.6.0 shim on PATH, exit 2 with the stale 1.5.3 one. AGENTS.md previously recorded this target as broken; that note is corrected.

## The guardrail

`find_legacy_python_invocations` scans **773 markdown, shell, and python files** and fails on the four refused forms. Error-level rather than a warning, which only became possible because everything else in this PR is fixed.

Eighteen-plus self-test fixtures cover it: nine assert it fires, nine assert it stays quiet on `uv run --no-project`, `uv add`, `python3 -c`, `uv pip --directory`, prose prohibitions, dockerfile fences and the allow-marker. Mutation-tested in both markdown and shell.

Worth recording that the check caught its own worst bug during development: with unanchored patterns it flagged `uv run --no-project python fuzz.py` — the very form its advice recommends — because that string contains `python fuzz.py`. The compliant-prefix guard exists for that reason.

Self-test goes 45 → 74 assertions across four review rounds, and the floor was mutation-tested (dropping fixtures reports "ran only 49, below the floor of 56").

## Verification

- Validator: zero errors, 1002 files scanned, the same 8 skill-length warnings as before and no new ones.
- `prek run -a`: all 15 hooks pass.
- `make check` still fails only at the **pre-existing** `constant-time-analysis` toolchain gap (28 failed / 94 passed, byte-identical to the baseline recorded before this branch). That plugin needs cross-compilation toolchains this machine lacks and CI installs.
- One flaky unrelated failure seen once and not reproducible: `writing-lean-proofs`' eval self-test shells out to `claude`, so it depends on a live model call.

Twelve plugins bumped in both `plugin.json` and `marketplace.json` — minor where a skill went from unusable to working, patch for install-form corrections.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -20,7 +20,7 @@
     },
     {
       "name": "building-secure-contracts",
-      "version": "1.1.2",
+      "version": "1.1.3",
       "description": "Comprehensive smart contract security toolkit based on Trail of Bits' Building Secure Contracts framework. Includes vulnerability scanners for 6 blockchains and 5 development guideline assistants.",
       "author": {
         "name": "Omar Inuwa && Pawe\u0142 P\u0142atek"
@@ -47,7 +47,7 @@
     },
     {
       "name": "constant-time-analysis",
-      "version": "0.2.0",
+      "version": "0.2.1",
       "description": "Detect compiler-induced timing side-channels in cryptographic code",
       "author": {
         "name": "Scott Arciszewski",
@@ -57,7 +57,7 @@
     },
     {
       "name": "culture-index",
-      "version": "1.1.1",
+      "version": "1.1.2",
       "description": "Interprets Culture Index survey results for individuals and teams",
       "author": {
         "name": "Dan Guido"
@@ -147,7 +147,7 @@
     },
     {
       "name": "semgrep-rule-creator",
-      "version": "1.2.2",
+      "version": "1.2.3",
       "description": "Create custom Semgrep rules for detecting bug patterns and security vulnerabilities",
       "author": {
         "name": "Maciej Domanski"
@@ -177,7 +177,7 @@
     },
     {
       "name": "static-analysis",
-      "version": "1.3.1",
+      "version": "1.3.2",
       "description": "Static analysis toolkit with CodeQL, Semgrep, and SARIF parsing for security vulnerability detection",
       "author": {
         "name": "Axel Mierczuk & Pawe\u0142 P\u0142atek"
@@ -195,7 +195,7 @@
     },
     {
       "name": "testing-handbook-skills",
-      "version": "1.0.2",
+      "version": "1.1.0",
       "description": "Skills from the Trail of Bits Application Security Testing Handbook (appsec.guide)",
       "author": {
         "name": "Pawe\u0142 P\u0142atek"
@@ -204,7 +204,7 @@
     },
     {
       "name": "trailmark",
-      "version": "0.10.1",
+      "version": "0.11.0",
       "description": "Builds multi-language source and binary code graphs for security analysis: call graphs, attack surface mapping, blast radius, taint propagation, complexity hotspots, entry point enumeration, proxy/unresolved-call tracking, type/reference analysis, and structural diffs. Creates bounded, graph-informed source packets for delegating focused work to constrained subagents. Generates Mermaid diagrams, runs graph-informed mutation testing triage (genotoxic), generates mutation-driven test vectors (vector-forge), extracts crypto protocol message flows, converts Mermaid diagrams to ProVerif models, projects SARIF/weAudit/binary findings onto code graphs, triages single findings with graph evidence, gates branch diffs for structural review regressions, and expands seed findings into variant-neighborhood candidates. Use when analyzing call paths, slicing source context for smaller models, mapping attack surface, visualizing code architecture, triaging survived mutants, generating cryptographic test vectors, diagramming crypto protocols, formally verifying protocols, augmenting audits with static analysis findings, deciding whether one candidate issue is reachable, reviewing graph-level PR risk, or seeding variant analysis.",
       "author": {
         "name": "Scott Arciszewski",
@@ -214,7 +214,7 @@
     },
     {
       "name": "variant-analysis",
-      "version": "2.0.1",
+      "version": "2.0.2",
       "description": "Find similar vulnerabilities and bugs across codebases using pattern-based analysis",
       "author": {
         "name": "Axel Mierczuk"
@@ -223,7 +223,7 @@
     },
     {
       "name": "c-review",
-      "version": "1.2.0",
+      "version": "1.3.0",
       "description": "Comprehensive C/C++ security code review with specialized bug-finding agents covering memory safety, type safety, concurrency, and Linux/Windows userspace-specific issues",
       "author": {
         "name": "Pawe\u0142 P\u0142atek"
@@ -232,7 +232,7 @@
     },
     {
       "name": "rust-review",
-      "version": "1.0.0",
+      "version": "1.1.0",
       "description": "Comprehensive Rust security code review with specialized bug-finding agents covering the safe/unsafe boundary, memory safety in unsafe blocks, concurrency, panic-induced DoS, recursion-induced stack overflow, FFI, and async runtime hazards",
       "author": {
         "name": "Andrea Cappa (Aptos Labs) & Pawe\u0142 P\u0142atek (Trail of Bits)"
@@ -272,7 +272,7 @@
     },
     {
       "name": "yara-authoring",
-      "version": "2.1.0",
+      "version": "2.1.1",
       "description": "YARA-X detection rule authoring with linting and quality analysis",
       "author": {
         "name": "Trail of Bits",
@@ -312,7 +312,7 @@
     },
     {
       "name": "zeroize-audit",
-      "version": "0.1.2",
+      "version": "0.2.0",
       "description": "Detects missing or compiler-optimized zeroization of sensitive data with assembly and control-flow analysis",
       "author": {
         "name": "Trail of Bits",
--- a/.github/scripts/validate_plugin_metadata.py
+++ b/.github/scripts/validate_plugin_metadata.py
@@ -105,21 +105,27 @@ LEGACY_PYTHON_PATTERNS = (
         "uses the legacy `uv pip` interface; use `uv add`, `uv sync`, or `uv tool install`",
     ),
     (
-        # Leading short flags are stepped over (`python3 -u foo.py` is refused too),
-        # except -c/-m, whose argument is not a script path.
-        re.compile(r"\bpython3?\s+(?:-(?![cm]\b)[A-Za-z]\S*\s+)*(?!-)\S*\.py\b"),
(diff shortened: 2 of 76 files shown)


dguido:
Acted on the review in `1c55014`. This was the most useful review of the series — the "changed *where* a package lands" framing caught a class of bug my own review pass missed entirely. Dispositions:

**P2 trailmark — fixed.** You were right that `uv tool install` cannot serve the library imports, and the failure mode you described (the MANDATORY retry block naming the exact ModuleNotFoundError it would loop on) is real. The CLI install stays `uv tool install`; the import-snippet wrappers are now `uv run --with trailmark python -`, which serves imports without touching the audited repo.

**P2 `uv add` in the target repo — fixed for sarif-parsing, kept for atheris, deliberately.** sarif-parsing's rows, ijson comment and jsonschema example now use `uv run --with <pkg>`, and the checker's advice string leads with that form. atheris keeps `uv add`: the fuzzing harness directory is the user's own project — the flow now opens with `uv init --bare` to make that explicit — so manifest writes there are intended, not churn.

**P2 `uv sync` / ct-analyzer — fixed with a different remedy.** Rather than prefixing sixteen usage lines with `uv run`, the install is now `uv tool install .`, which puts the console script on PATH so every existing bare invocation keeps working. Verified end to end on this machine: installed, `--help` ran, uninstalled.

**P2 checker narrower than shim — fixed, and your live instance confirmed.** `python3 --version` at constant-time-analysis README:196 is indeed refused by the 1.6.0 shim (reproduced directly). The pattern set now covers flags-only invocations, leading short flags (`python3 -u foo.py`), and every `uv pip` subcommand, with `-t` added to the tool-managed allowance.

**P2 `.py` scanning — fixed.** Extended the scan; it found your full list plus two you didn't have (`opencv_extractor.py:141`, `generate_poc.py:14`). All ten-plus usage strings rewritten, including the `--help` of the planner this PR exists to fix.

**P3s — all fixed.** Relative-path exemption (your checkout-under-`~/tests` scenario no longer bypasses the scan); marker scope now ends at a blank line, and `quality-assessment.md` gained the second marker that scoping made necessary; yarGen runs under `--with-requirements requirements.txt`; sarif-tools' scripting row is importable; zeroize's preflight got `which python3` back; cbor2's source build persists via `[tool.uv] no-binary-package` (field verified against uv's accepted-settings list — it errors on unknown fields, so this was cheap to prove).

**P3 `instead of` exempting whole lines — acknowledged, kept.** It's a prohibition heuristic with a known false-negative direction; the alternative is parsing sentence structure. Since a missed violation here surfaces the moment anyone runs the command, I'll take the simpler rule.

**P4s.** The 12-vs-13 inconsistency was already gone in `4b306d5` (the narrative comments were trimmed). Makefile note rewritten. The `-m pip` fixture now uses `pip download` so pattern 3 is actually exercised. `python_docs_scanned` prints in the success line: current run is `0 findings over 773 files`. The count discrepancy resolves in favor of neither of us — my 698 predates the `.py` extension and your 688 was reconstructed with find/grep; the executed number now is 773 with `.py` included. Function name in the PR body: you're right, it's `find_legacy_python_invocations`; body corrected below rather than renaming a function to match prose. Caracal CI example now carries a cache/pin note. `cargo install --git https://github.com/crytic/caracal --profile release --force` is verbatim from upstream's README "Building from git" section, fetched during the original change.

Self-test went 56 → 63; every new pattern and exemption is fixture-covered, and each was mutation-probed against the real tree (plant, confirm error, remove, confirm clean).

dguido:
Second round addressed in `76d276d`. Dispositions, numbered as in the review:

**1 (P2) — fixed.** You were right that this PR traded one broken population for another: Phase 1 (Prerequisites) in both review skills now probes `command -v uv` and aborts with install guidance, and the four converted shell suites carry the same guard with a message instead of a bare 127. zeroize-audit's skill-level preflight already listed `uv --version`, which covers its agents.

**2, 3, 4, 5 — fixed.** pipx and the non-install pip subcommands get named-subcommand patterns; the variable/path-script pattern landed and immediately caught one live instance (a codeql test stub that fakes uv itself, now allow-marked with its reason); `finditer` closes the masking case; prohibitions now test the text *before* the match, so your "instead of the tarball" counterexample fires while trailmark's "Do NOT run" stays exempt. The masking fix surfaced a subtlety worth recording: an *allowed* `uv pip install --directory …` contains a literal `pip install`, so a pip-family match directly after `uv ` now defers to the uv-pip verdict rather than double-reporting with the wrong advice.

**6, 7 — fixed.** The jsonschema line no longer names a nonexistent script and the table's run-forms show a concrete `script.py`. On the table itself I kept run-forms rather than your suggested `uv add pysarif`/`uv tool install sarif-tools`: round one of this review argued, correctly in my view, that `uv add` in the operator's CWD writes to the audited repo's manifest, and `from sarif import loader` needs importability a tool env cannot give. The two rounds pointed in opposite directions here; placement wins. culture-index's two messages now agree and name `uv run --project` on the scripts directory.

**8-18 (P4s) — all fixed except two accepted.** Stale comments corrected (`only .md`, "four refused forms", variant-analysis's stdin claim); pattern order gives `uv pip` lines the right advice; the flag allowance matches whole flags (`--target-dir` fires, trailing `-t /tmp` is quiet — both fixtures now); yarGen regains `--no-project` plus a `cd` into the checkout; generate_poc's `requires-python` is `>=3.10` — good catch, `str | None` in a signature is a def-time TypeError on 3.9 and uv's interpreter selection made it reachable; the GitLab snippet exports `~/.local/bin`; the generator's install section stops prescribing a venv its own runner never uses; PR body numbers refreshed below.

Accepted as-is: **9's** tests/evals `.py` exemption stays — test files legitimately quote refused commands as expectations, and the exemption is now correctly relative to `plugins/`; **14's** widening direction is covered by the existing allow-marker fixture (marker + violation on the contiguous next line, quiet) paired with the blank-line fixture (fires), which pins both edges of the scope.

Self-test 63 → 71. Full scan still 0 findings over 773 files.

dguido:
Third round addressed in `ab79a7a`.

**P2 — fixed, and fair: it was the residue of your round-two finding, applied to the siblings but not the flagship.** trailmark/SKILL.md's install block now says what each remedy is for — `uv tool install` provides the CLI only; import errors in snippets are cured by `uv run --with trailmark python -`, not by installation — and the other five library-first docs carry the same one-line annotation beside their install command.

**Declined with a transcript — `uv run python3` as non-compliant.** Tested with the 1.6.0 shim first on PATH, outside any project: `uv run --no-project python3 x.py` prints `ran ok`, as do the `python` and bare-script forms, and the Makefile's own `uv run --no-project python3 …` self-test passes. uv prepends its environment's bin directory to the child PATH, so `python3` resolves to a real interpreter there; the shim header's "resolves back to this shim" describes bare suggestion text outside `uv run`, not commands under it. That also disposes of the Makefile and `intercept-github-curl.bats:103` corollaries.

**Declined with a citation — zeroize's missing uv prerequisite.** `prompts/system.md` lists `uvx` under required tools for C/C++ and `uv` for Rust, and the quick check runs `uv --version`; the C/C++ `which` line now names `uv` explicitly too.

**Fixed:** the ct-analyzer probe — good catch, it subprocessed `python3 --version`, the one refused form, and would report "Python is not available" under the shims; it now probes `sys.executable` and returns True with the shim on PATH (verified). Flag step-over now handles long and value-taking flags (`-W ignore`, `--verbose`), matching the shim's two-slot consumption, with fixtures. A bare `allow-legacy-python:` no longer exempts — the reason is enforced. The ten bare `uv run {baseDir}` sites in semgrep and culture-index gained `--no-project`, and culture-index's remediation strings now name that exact runnable command (you were right that `--project` named a mechanism nothing uses). pip gains `cache`/`config` and the comment now says the list is deliberately a subset. All the P4 text corrections landed (phase range, run_fixtures' false PEP 723 claim, yarGen cd notes, tool-bin PATH exports and notes at all four `uv tool install` sites, the sarif column and the invented filename).

**One find of our own while running your fixes:** a stray `.venv` this session's uv probes left inside constant-time-analysis turned the path scan red locally — invisible to CI, which never has env dirs. Both filesystem scans now skip `.venv`/`node_modules`-style directories, keeping local and CI runs equal.

Self-test 71 → 74. Full scan: 0 findings over 773 files.

PR #267, opened 2026-08-19
zeroize-audit: give SKILL.md a run entry point and link the Rust patterns reference

### LINKED ISSUE #261: zeroize-audit: SKILL.md never tells the model how to run the plugin; 867-line reference is orphaned
Two related gaps found in Phase 1 triage.

**1. No entry point.** `skills/zeroize-audit/SKILL.md` runs When-to-Use → When-NOT → Purpose → Scope → Inputs → Prerequisites → … and never instructs the model to read `prompts/task.md` and execute its phase loop. The only mention is a parenthetical around line 154. So the plugin's entire orchestration — the thing it exists to do — has no reachable instruction telling the model to start it.

**2. Orphaned reference.** `references/rust-zeroization-patterns.md` (867 lines, 41 named Rust zeroization anti-patterns across four sections) has **zero inbound references**: `rg -l 'rust-zeroization-patterns' plugins/zeroize-audit` excluding the file itself returns nothing. It ships to every user and is never pointed at.

**Fix:** add a "How to run" section to SKILL.md that maps a natural-language request onto the `prompts/task.md` phase loop; link `rust-zeroization-patterns.md` from SKILL.md and from the Rust analyzer agents (`agents/2b-rust-source-analyzer.md`, `agents/3b-rust-compiler-analyzer.md` — confirm exact filenames on disk first). Done when the reference has at least one inbound link and a fresh read of SKILL.md leads to executing the task loop.

dguido:
Fixes #261

## What changed

`plugins/zeroize-audit/skills/zeroize-audit/SKILL.md` documented what the skill detects but never told the model how to start a run. `prompts/task.md` holds the phase loop the entire pipeline hangs off, and it appeared only as a parenthetical in the Agent Architecture caption. A new **How to Run** section, placed directly after When NOT to Use, maps a natural-language request onto the flow that actually exists:

1. collect inputs (`path`, plus `compile_db` and/or `cargo_manifest`, else preflight stops the run)
2. read `{baseDir}/prompts/task.md` for the orchestrator role, and `{baseDir}/prompts/system.md` for the working-directory layout and agent error protocol
3. execute the phase loop, reading `{baseDir}/workflows/phase-{N}-{name}.md` before each phase and honoring its skip conditions
4. Phase 8: read and return `{workdir}/report/final-report.md`

plus the `orchestrator-state.json` resume path from task.md's Recovery section. Every step is taken from `prompts/task.md` and the workflow files, not invented.

`references/rust-zeroization-patterns.md` (867 lines, 40 named patterns) had zero inbound references. It is now linked three times, each pointing at the section covering the patterns that reader's scripts match:

- **SKILL.md**, Detection Strategy: the whole file, keyed to the detecting script per section
- **`agents/2b-rust-source-analyzer.md`**: Section A (`A1`–`A12`) after the `semantic_audit.py` detection list, Section B (`B1`–`B10`) after the `find_dangerous_apis.py` list
- **`agents/3b-rust-compiler-analyzer.md`**: Section C (`C-MIR1`–`C-MIR3`, `C-IR1`–`C-IR5`, `C-ASM1`–`C-ASM4`) mapped to Steps 2, 4, and 4b, plus Section D as a coverage gap a clean run does not rule out

Both agent filenames were confirmed on disk before linking. Version bumped 0.2.0 → 0.3.0 in `plugin.json` and `.claude-plugin/marketplace.json` (surgical edits to the two version lines).

The reference says 40 named patterns, not the 41 in the issue: `rg -c '^### '` over the file returns 40 (A: 12, B: 10, C: 12, D: 6). The prose uses the verified number.

A second commit corrects how much coverage the link text claimed. Section C documents 12 patterns while the three scripts match roughly twenty: `check_mir_patterns.py` defines 8 detectors against 3 entries, and `check_llvm_patterns.py` emits secret-return-value and by-value-aggregate classes with no entry at all. The text now calls Section C a subset and tells the agent that a pattern with no entry is still a valid finding carrying the script's own evidence, because otherwise an absent entry reads as "not a real pattern" and the finding goes unreported. Section A is likewise not a 1:1 map of the 2b bullet list (the missing-`zeroize`-dependency check has no entry; `A6`'s `ManuallyDrop<T>` field has no bullet), 34 of the 40 patterns are detected while Section D's 6 are known gaps, and C-ASM3/C-ASM4 carry no snippet. That commit also fills in the `max_tus` (50) and `mcp_timeout_ms` (10000) defaults from `schemas/input.json` that the new step 1 depends on, and adds the three `references/` files missing from the README index.

## Verification

Inbound references to the formerly orphaned file:

```
$ rg -l 'rust-zeroization-patterns' plugins/zeroize-audit
plugins/zeroize-audit/skills/zeroize-audit/SKILL.md
plugins/zeroize-audit/agents/3b-rust-compiler-analyzer.md
plugins/zeroize-audit/agents/2b-rust-source-analyzer.md
```

(the reference file itself does not self-match, so all three are new inbound links)

Validator:

```
$ uv run --no-project python .github/scripts/validate_plugin_metadata.py .
Checking 41 plugin(s)

8 warning(s) — these do not fail the build:

  ! testing-handbook-skills: skills/aflpp/SKILL.md is 629 lines, over the 500 limit
  ! testing-handbook-skills: skills/atheris/SKILL.md is 519 lines, over the 500 limit
  ! testing-handbook-skills: skills/constant-time-testing/SKILL.md is 507 lines, over the 500 limit
  ! testing-handbook-skills: skills/coverage-analysis/SKILL.md is 606 lines, over the 500 limit
  ! testing-handbook-skills: skills/harness-writing/SKILL.md is 614 lines, over the 500 limit
  ! testing-handbook-skills: skills/libafl/SKILL.md is 625 lines, over the 500 limit
  ! testing-handbook-skills: skills/libfuzzer/SKILL.md is 795 lines, over the 500 limit
  ! testing-handbook-skills: skills/wycheproof/SKILL.md is 533 lines, over the 500 limit

✓ no errors (395 references resolved, 989 files scanned for hardcoded paths, 763 for legacy python invocations)
```

Zero errors. All 8 warnings are pre-existing SKILL.md length warnings in `testing-handbook-skills`; none are dangling references and none are in `zeroize-audit`. The new links use the `{baseDir}/references/...` form this plugin already uses for `detection-strategy.md` and `poc-generation.md`; every path they name exists on disk (`ls` confirms `references/rust-zeroization-patterns.md`, `prompts/task.md`, `prompts/system.md`, `schemas/input.json`). SKILL.md is 375 lines, still under the 500-line limit.

Other checks: `make self-test` (74 assertions), `make lint`, `make shell`, and `make validate` all pass. `plugins/zeroize-audit/tests/rust-regression/run_smoke.sh` passes.

Two pre-existing failures in `make check` are unrelated to this diff, which touches only Markdown and two version strings — both were confirmed by stashing the changes and re-running on a clean tree:

- `make bats`: 4 failures in `plugins/modern-python/hooks/shims/uv-shim.bats` (`allows uv pip when --directory/--project/--target/-t says a tool owns the environment`), identical count before and after the change. Local-only: CI's Shell (bats) job passes on this branch, so the difference is the local `uv` version, not the repo.
- `make python-tests`: `plugins/constant-time-analysis` fails on a missing Java toolchain (`javac/javap`) and a missing Rust `std` for the test target — environmental, not code

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -312,7 +312,7 @@
     },
     {
       "name": "zeroize-audit",
-      "version": "0.2.0",
+      "version": "0.3.0",
       "description": "Detects missing or compiler-optimized zeroization of sensitive data with assembly and control-flow analysis",
       "author": {
         "name": "Trail of Bits",
--- a/plugins/zeroize-audit/.claude-plugin/plugin.json
+++ b/plugins/zeroize-audit/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "zeroize-audit",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "description": "Detects missing or compiler-optimized zeroization of sensitive data with assembly and control-flow analysis",
   "author": {
     "name": "Trail of Bits",
--- a/plugins/zeroize-audit/README.md
+++ b/plugins/zeroize-audit/README.md
@@ -295,5 +295,8 @@ Supported categories (C/C++ PoCs): `MISSING_SOURCE_ZEROIZE`, `OPTIMIZED_AWAY_ZER
 ## References
 
 - `skills/zeroize-audit/references/compile-commands.md`
+- `skills/zeroize-audit/references/detection-strategy.md`
 - `skills/zeroize-audit/references/ir-analysis.md`
 - `skills/zeroize-audit/references/mcp-analysis.md`
+- `skills/zeroize-audit/references/poc-generation.md`
+- `skills/zeroize-audit/references/rust-zeroization-patterns.md`
--- a/plugins/zeroize-audit/agents/2b-rust-source-analyzer.md
+++ b/plugins/zeroize-audit/agents/2b-rust-source-analyzer.md
@@ -69,7 +69,9 @@ This detects:
 - `#[derive(Serialize)]` on sensitive type → `SECRET_COPY` (low)
 - No `zeroize` crate in `Cargo.toml` → `MISSING_SOURCE_ZEROIZE` (low)
 
-Section A of `{baseDir}/references/rust-zeroization-patterns.md` documents most of these as `A1`–`A12`, each with a minimal reproducing snippet and the recommended fix. Read the entry for a pattern before writing its finding `detail`, and use its snippet to judge whether a flagged type really matches or is a false positive.
+Section A of `{baseDir}/references/rust-zeroization-patterns.md` documents most of these as `A1`–`A12`, each with a minimal reproducing snippet and the recommended fix. Read the entry for a pattern before writing its finding `detail`, and use its snippet to judge how closely the flagged type matches.
+
+Never drop or downgrade a finding because it reads as a poor match for the entry. A weak match is a `needs_review` finding whose `evidence` says how the code differs from the entry, not an omission: Step 5 combines both arrays in full, and a finding left out here is unrecoverable from any artifact the run produces.
 
 The mapping is not one-to-one. The missing-`zeroize`-dependency check has no Section A entry, and `A6` (a `ManuallyDrop<T>` struct field) covers a pattern the list above does not name. Where a check has no entry, write the `detail` and the fix from that check's own output — borrowing a neighbouring entry's snippet produces a finding that describes the wrong flaw.
 
--- a/plugins/zeroize-audit/agents/2b-rust-source-analyzer.md
+++ b/plugins/zeroize-audit/agents/2b-rust-source-analyzer.md
@@ -69,6 +69,8 @@ This detects:
 - `#[derive(Serialize)]` on sensitive type → `SECRET_COPY` (low)
 - No `zeroize` crate in `Cargo.toml` → `MISSING_SOURCE_ZEROIZE` (low)
 
+Section A of `{baseDir}/references/rust-zeroization-patterns.md` documents each of these as `A1`–`A12`, with a minimal reproducing snippet and the recommended fix. Read the entry for a pattern before writing its finding `detail`, and use its snippet to judge whether a flagged type really matches or is a false positive.
+
 If the script is missing or fails: write a status-bearing error object to the output file and continue:
 
 ```json
@@ -102,6 +104,8 @@ This detects:
 - `slice::from_raw_parts` → `SECRET_COPY` (medium)
 - `mem::take` → `MISSING_SOURCE_ZEROIZE` (medium)
 - async fn with secret-named local + `.await` → `NOT_ON_ALL_PATHS` (high)
+
+Section B of `{baseDir}/references/rust-zeroization-patterns.md` covers these as `B1`–`B10`. Each entry explains why the API defeats zeroization, which matters here because this scanner is token-based: it cannot tell `mem::take` used to steal a secret from `mem::take` used on an unrelated buffer.
 
 Findings without sensitive names in ±15 surrounding lines are downgraded to `needs_review`.
 
--- a/plugins/zeroize-audit/agents/2b-rust-source-analyzer.md
+++ b/plugins/zeroize-audit/agents/2b-rust-source-analyzer.md
@@ -69,7 +69,9 @@ This detects:
 - `#[derive(Serialize)]` on sensitive type → `SECRET_COPY` (low)
 - No `zeroize` crate in `Cargo.toml` → `MISSING_SOURCE_ZEROIZE` (low)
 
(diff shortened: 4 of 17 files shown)


dguido:
Checked every claim against the scripts before touching anything. The four about my own text are correct and are fixed in c17698b.

**P2 (Section C not exhaustive) — fixed.** Verified: `check_mir_patterns.py` defines 8 detectors (`detect_drop_before_storagedead`, `detect_resume_with_live_secrets`, `detect_aggregate_move_non_zeroizing`, `detect_closure_capture_secret`, `detect_drop_glue_without_zeroize`, `detect_ffi_call_with_secret`, `detect_yield_with_live_secret`, `detect_result_err_path_with_secret`) against 3 Section C entries, and `check_llvm_patterns.py:analyze()` emits class 6 (secret return values → `REGISTER_SPILL`, line 361) and class 7 (by-value aggregate args → `SECRET_COPY`, line 375) with no entry. This was the one worth catching: the old wording let an absent entry read as "not a real pattern", which is the silent-recall-collapse failure `AGENTS.md` warns about. The agent now reads "Section C is a subset, not an index" and "Never drop or downgrade a finding because the reference does not describe it."

**P3 (Section A mapping) — fixed.** Confirmed: `rg 'Cargo.toml'` over the reference returns nothing, so the missing-dependency check has no entry, and `ManuallyDrop<T>` appears in the 2b file only as Step 3's `ManuallyDrop::new`, never as A6's struct-field pattern. The counts coincide at 12; the contents do not. The text now says so and tells the agent to write the detail from the check's own output rather than borrowing a neighbour's snippet.

**P4 (40 vs 34) — fixed.** Sections A-C hold 34 detected patterns; Section D's 6 are gaps. SKILL.md no longer calls all 40 "what the tooling looks for" and now says to treat Section D patterns as unaudited rather than clean.

**P4 (C-ASM3/C-ASM4 snippets) — fixed.** Both go Tool → Why source is blind → Detection with no code block, so "each entry gives a reproducing snippet" is now "most also give". The Detection-line claim held for all 12, so that part stands.

**P4 (README index, blank defaults) — fixed.** Added the three missing `references/` entries, and filled in `max_tus: 50` / `mcp_timeout_ms: 10000` from `schemas/input.json`, since the new step 1 leans on that column.

**P2 (bare agent names vs `<plugin>:<agent>`) — not fixed, deliberately.** The finding is right and I could not verify otherwise: all 11 spawn sites across the 8 workflows use bare names, nothing outside the metadata files mentions `zeroize-audit:`, and no literal `subagent_type` key exists for the validator's regexes to catch. But this is a behavioral change to how the pipeline dispatches, it predates this branch, and I cannot test that a namespaced value actually dispatches from here — a docs PR is the wrong vehicle for guessing at it. Worth its own issue against the workflow files. Fair point that this branch makes it the documented path; it does not make it more broken, it makes the breakage reachable.

**P3 (Section D note lands in an unread `notes.md`) — not fixed.** Verified that `4-report-assembler.md` Step 1 reads findings JSONs, `sensitive-objects.json`, `mcp-evidence/status.json`, and `preflight.json`, and no agent's `notes.md`. The routing you propose (a `needs_review` finding, or feeding gaps to the assembler) is a real improvement and also a pipeline change beyond this issue — and emitting a finding for every `Arc<T>` in a crate needs a false-positive budget I would not set blind. `notes.md` remains where the plugin's own agent protocol says per-agent observations go, so the instruction is consistent even though the artifact is not surfaced. Same recommendation: separate issue.

**P4 (`baseDir` labelled "Plugin base directory")** — agreed and pre-existing, no failure under the current flow since `phase-0-preflight.md:29` passes the token through. Noted, not touched.

dguido:
Second round of review findings addressed; verified each first.

- **`coverage-gaps.json` read in a mode that cannot write the report** — real. Step 1 runs in interim mode only, Analysis Coverage is Step 6 (final-only), so the file was read and discarded and the report still said "no coverage gaps were reported". Moved the read to the point of use.
- **Nothing instructed 3b to perform the survey** — real. The paragraph sat above Step 1 and the numbered steps never looked, while five of six D patterns are source constructs 3b has no method for. Added Step 6 with grep markers per pattern and a sensitive-type filter, required the file unconditionally (`[]` when empty, so "survey ran, found nothing" is distinguishable from "no survey"), and declared it in the phase-2 workflow, the always-write list, `system.md`, and `report_template.md`.
- **Namespacing left the validator disarmed** — real and the worst of the three. `` `subagent_type: zeroize-audit:X` `` matches neither extractor (both need a quote or backtick immediately after the colon), so `validate_subagent_dispatch` inspected **zero** sites here. Quoted the values: the validator now sees all 12, and reverting one to a bare name fails the build with `subagent_type '0-preflight' is not namespaced`.

kz-tob (approved):

PR #268, opened 2026-08-19
Close unterminated fences in cairo, algorand, and ton scanner skills

### LINKED ISSUE #260: building-secure-contracts: an automated split corrupted 3 of 4 chain scanners
The cairo, algorand, and ton vulnerability-scanner `SKILL.md` files each carry the same corruption signature from an automated content split: **three duplicate `## 5.` headings** and an **odd number of code fences**, so everything after the unterminated fence renders as a code block. The truncated section is "## 5. Example Output".

Measured this session (fence count via `rg -c '^\`\`\`'`, heading count via `rg -c '^## 5\.'`):

| file | `## 5.` headings | code fences |
|---|---|---|
| cairo-vulnerability-scanner | 3 | 19 (odd) |
| algorand-vulnerability-scanner | 3 | 13 (odd) |
| ton-vulnerability-scanner | 3 | 21 (odd) |
| **solana-vulnerability-scanner** | **1** | **24 (even)** |

solana is the clean sibling — same plugin, same structure, uncorrupted — so it's the template for what the other three should look like.

**Impact:** roughly the back half of each of the three skills is swallowed into a code block, so the model reads it as literal text rather than instructions. Invisible to CI: the validator does not check fence balance.

**Fix:** close/repair the fences, deduplicate the `## 5.` headings, renumber the trailing sections. Done when `rg -c '^\`\`\`'` is even and `rg -c '^## 5\.'` is 1 for each of the three files, and each renders cleanly.

dguido:
Fixes #260

## What was wrong

The cairo, algorand, and ton scanner skills each opened a code fence for the "Example Output" report and never closed it, so a long stretch of each file rendered as one code block. Under a CommonMark parser, only 8 of each file's 12 `## N.` heading lines rendered as headings before this PR. Each file also carried three H2 sections numbered `## 5.`, and the "Scanning Workflow" heading was glued to the preceding paragraph with no blank line.

`solana-vulnerability-scanner` is the fourth sibling, repaired the same way in #160 — but see the second commit: it shared one further defect with the other three.

## What changed

**Commit 1** — one closing fence per file at the end of the surviving example report, plus an H2 renumber to 1..12 in document order matching solana:

| section | before | after |
|---|---|---|
| Example Output | 5 | 5 |
| Vulnerability Patterns | 5 | 6 |
| Scanning Workflow | 5 | 7 |
| Reporting Format | 6 | 8 |
| Priority Guidelines | 7 | 9 |
| Testing Recommendations | 8 | 10 |
| Additional Resources | 9 | 11 |
| Quick Reference Checklist | 10 | 12 |

**Commit 2** — widened the Finding Template's outer fence from ` ```markdown ` to four backticks in **all four** files, solana included. The template nests triple-backtick code snippets, and per CommonMark a closing fence must be bare — so the first bare ` ``` ` after a nested snippet closed the outer block early and a later bare fence swallowed `## 9.` and `## 10.` as code. After commit 1 the files rendered 10 of 12 sections; after commit 2, 12 of 12.

No prose was added or rewritten. Version bumped 1.1.3 → 1.1.4 in `plugin.json` and `marketplace.json`.

## What could not be recovered

The example reports are still truncated, and I left them that way rather than inventing replacement text:

- **cairo** retains only its banner line, `=== CAIRO/STARKNET VULNERABILITY SCAN RESULTS ===`. The project/file-count summary and every finding are gone.
- **algorand** and **ton** both print `Vulnerabilities Found: 2` but show one finding, and that finding stops mid-entry — algorand's has no recommendation, ton's has no code or issue text.

This content was already missing in the initial import (695119c), so there is no earlier revision to restore it from. Note that solana's `## 5. Example Output` is an empty section for the same reason, so an empty or partial example is the existing norm for this family rather than a new gap.

## Verification

An earlier version of this description quoted a hand-rolled render check that toggled fence state on every ` ``` ` line. That check was wrong in both directions — it reported "5 sections" pre-fix where CommonMark renders 8 heading lines, and "12 sections, 12 max" after commit 1 when CommonMark renders 10 — because a parity toggle cannot model the bare-closer rule. Its output has been removed from this description; do not use it as evidence.

The evidence now comes from an actual CommonMark parser (`markdown-it-py`), counting `## N.` lines that render as headings:

```
pre-fix (main):        algorand 8/12   cairo 8/12   ton 8/12
after commit 1:        algorand 10/12  cairo 10/12  ton 10/12  solana 10/12
after commit 2 (HEAD): algorand 12/12  cairo 12/12  ton 12/12  solana 12/12
```

Each file has exactly one `## 5.`, headings run 1..12 in document order, and fence-marker lines are balanced in all four files.

Plugin validator: zero errors (8 pre-existing line-length warnings, all in `testing-handbook-skills`, untouched here). `make self-test eval-self-tests lint shell validate` pass. Two `make check` stages fail on this machine for reasons unrelated to this branch, both confirmed to fail identically with these changes stashed: 4 `plugins/modern-python/hooks/shims/uv-shim.bats` cases (the local modern-python 1.5.3 shim intercepting the test's own `uv` calls), and `constant-time-analysis` Python tests that need a Java toolchain and a Rust cross-arch stdlib that are not installed here. `building-secure-contracts` ships no Python tests.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -20,7 +20,7 @@
     },
     {
       "name": "building-secure-contracts",
-      "version": "1.1.3",
+      "version": "1.1.4",
       "description": "Comprehensive smart contract security toolkit based on Trail of Bits' Building Secure Contracts framework. Includes vulnerability scanners for 6 blockchains and 5 development guideline assistants.",
       "author": {
         "name": "Omar Inuwa && Pawe\u0142 P\u0142atek"
--- a/plugins/building-secure-contracts/.claude-plugin/plugin.json
+++ b/plugins/building-secure-contracts/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "building-secure-contracts",
-  "version": "1.1.3",
+  "version": "1.1.4",
   "description": "Comprehensive smart contract security toolkit based on Trail of Bits' Building Secure Contracts framework. Includes vulnerability scanners for 6 blockchains and 5 development guideline assistants.",
   "author": {
     "name": "Omar Inuwa && Paweł Płatek",
--- a/plugins/building-secure-contracts/skills/algorand-vulnerability-scanner/SKILL.md
+++ b/plugins/building-secure-contracts/skills/algorand-vulnerability-scanner/SKILL.md
@@ -175,7 +175,7 @@ For atomic transaction groups:
 ## 8. Reporting Format
 
 ### Finding Template
-```markdown
+````markdown
 ## [SEVERITY] Vulnerability Name (e.g., Missing RekeyTo Validation)
 
 **Location**: `contract.teal:45-50` or `approval_program.py:withdraw()`
@@ -209,7 +209,7 @@ If(And(
 **References**:
 - building-secure-contracts/not-so-smart-contracts/algorand/rekeying
 - Tealer detector: `unprotected-rekey`
-```
+````
 
 ---
 
--- a/plugins/building-secure-contracts/skills/algorand-vulnerability-scanner/SKILL.md
+++ b/plugins/building-secure-contracts/skills/algorand-vulnerability-scanner/SKILL.md
@@ -87,11 +87,11 @@ Code:
 
 Issue: The contract doesn't validate the RekeyTo field, allowing attackers
 to change account authorization and bypass restrictions.
-
+```
 
 ---
 
-## 5. Vulnerability Patterns (11 Patterns)
+## 6. Vulnerability Patterns (11 Patterns)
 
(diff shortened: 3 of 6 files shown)


dguido:
Verified the fence finding with a CommonMark parser (`markdown-it-py`): all four scanners — algorand, cairo, ton, and the untouched **solana** template — rendered only 10 of 12 `## N.` sections. Sections 9 and 10 were swallowed as code because the Finding Template's nested triple-backtick snippet closes the outer ` ```markdown ` block early (a closing fence must be bare per CommonMark). A fence-line parity count can't see this, which is why the earlier render check reported 12/12.

Fixed by widening the outer Finding Template fence to four backticks in all four files (solana included, since it shares the defect). The parser now renders 12/12 everywhere. Diff is fence markers only — 2 lines per file, no content touched.

kz-tob (approved):

PR #269, opened 2026-08-19
Fix firebase-apk-scanner scanner.sh path and command handoff

### LINKED ISSUE #263: firebase-apk-scanner: {baseDir}/scanner.sh does not resolve; command file contradicts skill frontmatter
`scanner.sh` lives at the **plugin root** (`plugins/firebase-apk-scanner/scanner.sh`), but the skill refers to it as `{baseDir}/scanner.sh`, and `{baseDir}` is the **skill directory** (`skills/firebase-apk-scanner/`). So the skill's central action fails:

- `SKILL.md:67` — `{baseDir}/scanner.sh $ARGUMENTS`
- `SKILL.md:5` — `allowed-tools: Bash({baseDir}/scanner.sh:*) …`

Both resolve to a path that does not exist, so the one command the skill exists to run cannot be invoked, and its own tool allowance would not match it.

**Fix:** `git mv` scanner.sh into the skill directory, matching every other component's layout, so `{baseDir}/scanner.sh` resolves.

**Also:** `SKILL.md:6` sets `disable-model-invocation: true` while `commands/scan-apk.md:14` says "Invoke the `firebase-apk-scanner` skill …". A non-model-invocable skill driven by a delegating command is a fine pattern, but the pair should be coherent — reconcile the wording so the command is unambiguously the entry point.

dguido:
Fixes #263

## Problem

`scanner.sh` lived at the plugin root (`plugins/firebase-apk-scanner/scanner.sh`), but the skill invoked it as `{baseDir}/scanner.sh`, and `{baseDir}` denotes the skill directory (`skills/firebase-apk-scanner/`). So `SKILL.md:67` and the `allowed-tools` pattern on `SKILL.md:5` both pointed at a path that did not exist, and the skill's central action could not run.

Separately, the skill sets `disable-model-invocation: true`, which blocks the model from invoking it. The `/trailofbits:scan-apk` command body said "Invoke the `firebase-apk-scanner` skill", an instruction Claude Code refuses for a non-model-invocable skill, so the command was a dead end.

## Changes

- `git mv` scanner.sh into the skill directory so `{baseDir}/scanner.sh` resolves (rename recorded at 100%, history preserved). `SKILL.md` lines 5 and 67 already used `{baseDir}`, so they need no change.
- Rework `commands/scan-apk.md`. Keep the shape (a non-model-invocable skill fronted by a delegating command), but instead of invoking the blocked skill, the command now reads and follows the skill's `SKILL.md` workflow directly and resolves `{baseDir}` via `${CLAUDE_PLUGIN_ROOT}`. The command is the unambiguous entry point and the skill is not told to invoke itself.
- Fix README usage. The command is `/trailofbits:scan-apk` (it previously showed `/firebase-scan`, which matches no command), and the standalone script path is now `skills/firebase-apk-scanner/scanner.sh`.
- Bump version 2.1.1 -> 2.2.0 in `plugin.json` and `marketplace.json`.

scanner.sh contains no `python3` or `pip` invocations, so no `uv` conversion was needed.

## Verification

- After the move, `skills/firebase-apk-scanner/scanner.sh` exists and the plugin root no longer has `scanner.sh` (the commit shows the rename at 100%).
- shellcheck on the moved script is clean (exit 0); the pre-commit `shellcheck` and `shfmt` hooks both Passed.
- The resolved `{baseDir}/scanner.sh` executes rather than reporting "file not found": running it with no arguments prints the usage banner and exits 1.
- `uv run --no-project python .github/scripts/validate_plugin_metadata.py .` reports no errors (395 references resolved, 989 files scanned for hardcoded paths, 763 for legacy python invocations). The 8 warnings it prints are pre-existing and all in the unrelated `testing-handbook-skills` plugin.
- The pre-commit `validate plugin metadata` hook Passed on commit.


dguido:
Verified and fixed the zero-config finding — it was real, and the same false-clean class as the original issue.

`extract_firebase_config` always writes a config file with `[]` for every field it found nothing for, so the `CONFIG_FAILED` branch only fires if the write itself fails. An APK that decompiles but carries no Firebase config skipped every endpoint test on its own guard and was written `SECURE`: `Total APKs: 1 / Vulnerable: 0` in green, exit 0, zero endpoints probed.

Now detected explicitly — status `NO_CONFIG`, counted in `untested_apks`, reported on its own summary line, and a run where nothing was tested exits 1. Verified with a stub apktool decompiling to an empty tree: before, exit 0 / `SECURE`; after, exit 1 / `NO_CONFIG` / `untested_apks: 1`.

The remaining findings (unrestricted `allowed-tools` on the command, JSON built by string interpolation, timestamped output dir globbing) are pre-existing and outside this fix; worth their own issue.

kz-tob (approved):

PR #271, opened 2026-08-19
static-analysis: resolve SARIF severity from the rule, not just result.level

### LINKED ISSUE #262: static-analysis/sarif-parsing: severity read from result.level, which SARIF makes optional — CI gate passes on a failing repo
SARIF 2.1.0 makes `result.level` **optional**; when absent it is inherited from the matched rule's `defaultConfiguration.level` (joined by `ruleIndex`/`ruleId`). The sarif-parsing skill reads `result.level` directly everywhere — SKILL.md (~5 sites), `resources/jq-queries.md` (~10), `resources/sarif_helpers.py` (~4).

**Impact:** on CodeQL output, which routinely omits `result.level`, every result reads as its default and the skill's documented CI severity gate **exits 0 on a repo full of errors**. This is the "a checker that passes on the wrong input" class AGENTS.md calls the most expensive bug in the repo.

**Fix:** severity lookup must resolve the rule's `defaultConfiguration.level` via `ruleIndex`, falling back to `result.level` when present. Prove it with a fixture SARIF whose results omit `level` (CodeQL-style): the gate must go **red** on it after the fix, and that fixture becomes a regression test. Show a normal SARIF still passes, so it is not just inverted.

**Secondary (same PR):** `compute_fingerprint` in `sarif_helpers.py` hashes `Path(file_path).name` only, so the same rule at the same line in two different directories collides. Include the normalized directory.

dguido:
Fixes #262

`result.level` is optional in SARIF 2.1.0, and CodeQL never emits it: severity lives on the rule as `defaultConfiguration.level` and the result inherits it. `sarif-parsing` read `result.level` directly in `sarif_helpers.py`, `resources/jq-queries.md` and `SKILL.md`, so the severity gate this skill documents counted zero errors on a CodeQL run however many it found.

## What changed

- `resolve_level(result, run)` and `find_rule(result, run)` in `sarif_helpers.py`. The rule is joined by `ruleIndex`, falling back to matching `ruleId` against `rules[].id` for tools that omit the index. Resolution order is `kind != "fail"` → `"none"`, then `result.level`, then the rule's `defaultConfiguration.level`, then `"warning"`. `extract_findings()` uses it, so `filter_by_level()`, `sort_by_severity()` and `count_by_level()` are correct without further change.
- The `kind` clause keeps the fix from inverting: a compliance tool's passing records would otherwise inherit their rule's `error` and fail a clean build.
- Every jq query that filters on severity, in both the reference and the SKILL, now starts from one `LEVEL_FN` definition, including the GitHub Actions gate that was the reported symptom.
- `compute_fingerprint()` hashed `Path(file_path).name` only, so the same rule at the same line in two directories produced one fingerprint and `deduplicate()` discarded the second finding. It now hashes the whole normalized path (POSIX separators, `file://` stripped, percent-decoded).
- A negative `ruleIndex` is rejected in both resolvers. SARIF writes `ruleIndex: -1` for "no rule", and `$rules[-1]` in jq is the *last* element, so the jq function would otherwise label those results with whatever severity the final rule in the array carries. Python's `0 <= index < len(rules)` bound already covered it.
- Two fixtures plus `resources/test_sarif_helpers.py` (22 tests). The suite runs the documented jq gate itself, extracted from the markdown, so the docs cannot drift from the helper.
- SKILL.md gained a severity section and lost three code blocks that duplicated the shipped helper (which is how one bug came to live in five places); it is 496 lines, under the 500 limit it was already near.

## Before and after, same bytes

The fixture is CodeQL-shaped: no result carries a level, the rules do.

```
$ jq -c '.runs[0].results[] | {ruleId, ruleIndex, level}' codeql-no-level.sarif
{"ruleId":"py/sql-injection","ruleIndex":0,"level":null}
{"ruleId":"py/clear-text-logging-sensitive-data","ruleIndex":null,"level":null}
{"ruleId":"py/unused-import","ruleIndex":2,"level":null}

$ jq -c '.runs[0].tool.driver.rules[] | {id, level: .defaultConfiguration.level}' codeql-no-level.sarif
{"id":"py/sql-injection","level":"error"}
{"id":"py/clear-text-logging-sensitive-data","level":"warning"}
{"id":"py/unused-import","level":null}
```

The gate documented at HEAD, against that file:

```
$ jq '[.runs[].results[] | select(.level == "error")] | length' codeql-no-level.sarif
0
```

The gate documented on this branch, against the same file:

```
$ jq "$LEVEL_FN"'[.runs[] as $run | $run.results[] | select(level($run) == "error")] | length' codeql-no-level.sarif
1
```

Not an inversion. Ordinary SARIF with severity on the results reports the same both ways:

```
before: 1
after:  1
```

Resolved severity per result:

```
error	py/sql-injection
warning	py/clear-text-logging-sensitive-data
warning	py/unused-import
```

`sarif_helpers.py` at HEAD vs this branch, over both fixtures:

```
codeql-no-level.sarif      before: 0 error / 3 findings           after: 1 error / 3 findings
levels-on-results.sarif    before: 1 error / 2 findings           after: 1 error / 2 findings

before: auth/login.py:42 -> 8d0a933265fdb66e  admin/login.py:42 -> 8d0a933265fdb66e
before: deduplicate() keeps 1 of 2 findings
after: auth/login.py:42 -> 319cfecc6485b38b  admin/login.py:42 -> ebc16dabc9b71ff7
after: deduplicate() keeps 2 of 2 findings
```

## Verification

New suite and the plugin's existing ones:

```
$ cd plugins/static-analysis/skills/sarif-parsing/resources
$ uv run --no-project --with pytest python3 -m pytest -q --import-mode=importlib .
22 passed in 0.13s

$ make python-tests   # static-analysis directories
  → plugins/static-analysis/skills/codeql/scripts
148 passed in 10.93s
  → plugins/static-analysis/skills/sarif-parsing/resources
22 passed in 0.13s
  → plugins/static-analysis/skills/semgrep/scripts
32 passed in 4.08s

$ bash plugins/static-analysis/tests/run_codeql_build_tests.sh   → 44 assertions passed
$ bash plugins/static-analysis/tests/run_scan_tests.sh           → 78 passed, 0 failed
$ bash plugins/static-analysis/tests/run_workflow_tests.sh       → 62 assertions passed
```

Validator:

```
$ uv run --no-project python .github/scripts/validate_plugin_metadata.py .
Checking 41 plugin(s)
✓ no errors (395 references resolved, 1004 files scanned for hardcoded paths, 775 for legacy python invocations)
```

Lint on the changed Python:

```
$ uvx ruff@0.14.13 check --output-format=concise ...   All checks passed!
$ uvx ruff@0.14.13 format --check ...                  2 files already formatted
```

Mutation check, on a throwaway copy, that the new tests fail when the bug returns:

| mutation | result |
|---|---|
| `level=result.get("level", "warning")` in `extract_findings` | 3 failed, 15 passed |
| `compute_fingerprint` back to `Path(file_path).name` | 3 failed, 15 passed |
| one documented jq query reverted to `select(.level == "error")` | 1 failed, 18 passed |
| fixture given `level` on its results (suite going vacuous) | 4 failed, 15 passed |
| `>= 0` guard dropped from the documented jq function | 2 failed, 19 passed |
| GitHub Actions gate reverted to `select(.level == "error")` | 2 failed, 20 passed |

(Each ran against the suite as it stood when that mutation was applied, so the totals grow down the table.)

`make check` was run too. `python-tests` fails in `constant-time-analysis` on this machine only, from absent toolchains (no `javac`, no rust `std` for riscv64, no C headers), and `make lint`, `make shell`, `make bats` and `make validate` pass. Version bumped 1.3.2 → 1.4.0 in `plugin.json` and `marketplace.json`.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -177,7 +177,7 @@
     },
     {
       "name": "static-analysis",
-      "version": "1.3.2",
+      "version": "1.4.0",
       "description": "Static analysis toolkit with CodeQL, Semgrep, and SARIF parsing for security vulnerability detection",
       "author": {
         "name": "Axel Mierczuk & Pawe\u0142 P\u0142atek"
--- a/plugins/static-analysis/.claude-plugin/plugin.json
+++ b/plugins/static-analysis/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "static-analysis",
-  "version": "1.3.2",
+  "version": "1.4.0",
   "description": "Static analysis toolkit with CodeQL, Semgrep, and SARIF parsing for security vulnerability detection",
   "author": {
     "name": "Axel Mierczuk & Paweł Płatek"
--- a/plugins/static-analysis/README.md
+++ b/plugins/static-analysis/README.md
@@ -42,6 +42,7 @@ Use this plugin when you need to:
 
 ### SARIF Parsing
 - Understand SARIF 2.1.0 structure
+- Resolve a result's severity from the rule it inherits it from, which CodeQL relies on
 - Quick analysis using jq for CLI queries
 - Python scripting with pysarif and sarif-tools
 - Aggregate and deduplicate results from multiple files
--- a/plugins/static-analysis/skills/sarif-parsing/SKILL.md
+++ b/plugins/static-analysis/skills/sarif-parsing/SKILL.md
@@ -165,15 +165,12 @@ for run in sarif.runs:
     tool_name = run.tool.driver.name
     print(f"Tool: {tool_name}")
 
-    # Rule defaults, for the results that carry no level of their own
-    rule_levels = {
-        r.id: getattr(r.default_configuration, "level", None)
-        for r in (run.tool.driver.rules or [])
-    }
-
     for result in run.results:
-        level = result.level or rule_levels.get(result.rule_id) or "warning"
-        print(f"  [{level}] {result.rule_id}: {result.message.text}")
+        # pysarif fills a missing result.level with "warning", so .level here is NOT the
+        # rule-inherited severity: a CodeQL error (no level on the result, severity on the
+        # rule) reads as "warning". Gate severity with Strategy 1's level() or with
+        # resolve_level() in resources/sarif_helpers.py, which resolve it from the rule.
+        print(f"  {result.rule_id}: {result.message.text}")
 
         if result.locations:
             loc = result.locations[0].physical_location
--- a/plugins/static-analysis/skills/sarif-parsing/SKILL.md
+++ b/plugins/static-analysis/skills/sarif-parsing/SKILL.md
@@ -47,7 +47,8 @@ sarifLog
     │   └── extensions[] (plugins)
     ├── results[] (findings)
     │   ├── ruleId
-    │   ├── level (error/warning/note)
+    │   ├── ruleIndex (index into tool.driver.rules[])
+    │   ├── level (OPTIONAL, inherited from the rule when absent)
     │   ├── message.text
     │   ├── locations[]
     │   │   └── physicalLocation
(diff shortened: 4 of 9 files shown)


dguido:
Worked the three review findings; verified each against the branch before touching anything.

- **Negative `ruleIndex` in the jq resolver** — already fixed. The bot reviewed the first commit; commit `04512f1` had already added the `and $r.ruleIndex >= 0` guard (both jq copies) plus `test_documented_jq_does_not_read_the_last_rule_for_rule_index_minus_one`. No change.
- **`resolve_level` mishandles explicit `"kind": null`** — real, fixed. `result.get("kind", "fail")` only defaults on a *missing* key, so an explicit `null` returned `"none"` and hid a real error, diverging from the jq gate's `// "fail"`. Changed to `(result.get("kind") or "fail")` and added a null/fail regression test. Mutation-checked: the test goes red (`none` vs `error`) on the old form.
- **pysarif Strategy 2 example** — real, and worse than flagged. pysarif fills a missing `result.level` with `"warning"`, so `result.level or rule_levels.get(...)` never reached the rule fallback — the rule-inheritance was dead code and a CodeQL error printed as `[warning]`. Dropped the dead dict and pointed severity gating at Strategy 1's `level()` / `resolve_level()`, which resolve from the rule.

24 tests pass.

PR #274, opened 2026-08-25
gh-cli: intercept GitHub fetches from MCP fetch tools

kz-tob:
## What

The gh-cli fetch hook only matched `WebFetch`. A GitHub URL fetched through an MCP fetch tool bypassed it entirely, so the plugin's premise — no unauthenticated GitHub access — did not hold for the fetch path many setups actually prefer.

The matcher is now `WebFetch|mcp__.*[Ff]etch`, which covers `mcp__exa__web_fetch_exa` and equivalents from other MCP servers.

A matcher alone is not enough. `WebFetch` passes a single `url`; MCP fetch tools pass a `urls` **array** and batch several pages into one call. The hook now reads both shapes:

```sh
urls=$(jq -r '[.tool_input.url?, (.tool_input.urls? | arrays | .[])]
  | map(select(type == "string" and . != ""))
  | .[]' 2>/dev/null) || exit 0
```

`arrays` drops a non-array `urls` instead of erroring; the `map(select(...))` drops nulls, numbers, and empty strings.

A tool call is atomic, so one GitHub URL anywhere in a batch denies the whole call, and each offending URL is labeled with its own suggestion so the model can tell which one was the problem. Single-URL calls keep the bare message they had.

The URL classification moved into `suggest_api`, `suggest_raw`, and `suggest_github_com` behind a `suggest_for_url` dispatcher, which is what lets it run per URL in a loop. That also collapses eight verbatim copies of the clone-hint string into one helper. Behavior is unchanged — `blob` and `tree` merge into one branch because they emitted identical text.

Also updated: the plugin README (matcher, both payload shapes, and the table now lists the `pull`, `issues`, `releases`, and `gist` patterns the hook already handled but never documented), the `SKILL.md` description, the root README row, and the version `1.5.1` → `1.6.0`.

## Testing

- 101/101 gh-cli bats tests pass — 32 pre-existing fetch, 10 new, 59 curl/session/shim.
- The 10 new tests were run against the old hook from `HEAD`: the 6 asserting new behavior fail there ("Expected deny JSON output, got empty"), and the 4 pass-through ones pass on both. The tests measure the fix rather than just accompanying it.
- `make lint shell validate` clean.
- Checked against a realistic Exa payload: `["https://docs.astral.sh/uv/", "https://raw.githubusercontent.com/astral-sh/uv/main/README.md"]` is denied with the offending URL named and a `gh repo clone astral-sh/uv` suggestion; the same batch without the GitHub URL passes through silently.

Two pre-existing failures on clean `main`, unrelated to this branch and confirmed by stashing: `make eval-self-tests` fails with 14 assertions in `property-based-testing/evals-extra/run.sh`, and `make python-tests` fails in `constant-time-analysis` on a machine without `javac` or a Rust `std` for the target.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -94,8 +94,8 @@
     },
     {
       "name": "gh-cli",
-      "version": "1.5.1",
-      "description": "Intercepts GitHub URL fetches and curl/wget commands, redirecting to the authenticated gh CLI.",
+      "version": "1.6.0",
+      "description": "Intercepts GitHub URL fetches (WebFetch and MCP fetch tools) and curl/wget commands, redirecting to the authenticated gh CLI.",
       "author": {
         "name": "William Tan",
         "email": "opensource@trailofbits.com",
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ cd /path/to/parent  # e.g., if repo is at ~/projects/skills, be in ~/projects
 |--------|-------------|
 | [code-improver](plugins/code-improver/) | Autonomous review-and-fix workflow over skills, plugins, or a branch, with a pluggable reviewer, findings ledger, escalation, and scope guard |
 | [devcontainer-setup](plugins/devcontainer-setup/) | Create pre-configured devcontainers with Claude Code and language-specific tooling |
-| [gh-cli](plugins/gh-cli/) | Intercept GitHub URL fetches and redirect to the authenticated `gh` CLI |
+| [gh-cli](plugins/gh-cli/) | Intercept GitHub URL fetches — `WebFetch`, MCP fetch tools, and curl/wget — and redirect to the authenticated `gh` CLI |
 | [git-cleanup](plugins/git-cleanup/) | Safely clean up git worktrees and local branches: a dynamic workflow gathers merge evidence and tries to refute its own delete recommendations, behind gated confirmation |
 | [goal-prompt](plugins/goal-prompt/) | Draft `/goal` commands for goal mode in Claude Code and Codex — verifiable completion conditions formatted to a copy-ready single line |
 | [github-triage](plugins/github-triage/) | Triage open GitHub issues and PRs: merge ready bot/approved PRs, review unreviewed ones via subagents, close resolved issues with cited comments, cross-link pending fixes, and score the rest with local-only priority and change-size estimates |
--- a/plugins/gh-cli/.claude-plugin/plugin.json
+++ b/plugins/gh-cli/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "gh-cli",
-  "version": "1.5.1",
-  "description": "Intercepts GitHub URL fetches and curl/wget commands, redirecting to the authenticated gh CLI.",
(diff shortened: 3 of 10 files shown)


kz-tob:
Thanks — worked through all of it. Pushed e4224bb.

**P1 (matcher may never match): checked, and it does match.** The docs settle the semantics: a matcher containing regex metacharacters is tested with unanchored `RegExp.test`, and `mcp__memory__.*` is written with a trailing `.*` because `mcp__memory` alone contains only exact-match characters and so is compared as a literal string — not because the match is anchored. Confirmed against the real name:

```
$ node -e 'console.log(new RegExp("WebFetch|mcp__.*[Ff]etch").test("mcp__exa__web_fetch_exa"))'
true
```

But the reason you couldn't settle it is the real finding, and it was right: **nothing tested the matcher, so a wrong one would have passed 101/101.** Added `hooks/matcher.bats` — 18 cases that pull the matcher out of `hooks.json` with jq and assert real tool names match or don't. Negative-controlled both ways: 4 fail against the old narrow matcher, and an empty matcher fails 14 via an `assert_matcher_nonempty` guard, since an empty pattern matches everything and would make every match assertion pass vacuously.

**P3 (scrape/crawl/extract uncovered): taken.** Matcher widened to `WebFetch|mcp__.*([Ff]etch|[Ss]crape|[Cc]rawl|[Ee]xtract)`, so Firecrawl scrape and Tavily extract are covered. Left browser navigation out. README no longer claims "any MCP fetch tool" and names what matches.

**P3 (string-valued `urls`): taken, and you were right that my test locked the bypass in.** Listed `urls` twice so the type filter keeps it as a bare string and `arrays` unpacks it as a list. Test flipped to `assert_deny`, plus a non-GitHub string case. Agreed on not going to `.. | strings` — `prompt` stays unscanned, and there's now a test asserting a prompt mentioning a GitHub URL doesn't false-deny.

**P4 (101/101 undercounts by 47): correct, my error.** `hooks/*.bats` missed `hooks/shims/gh-shim.bats`, which `find` picks up. The real figure was 148; with the matcher suite it's **166 passing, 0 failing**.

**P4 (README table): taken.** Added the `/contents/`, `/releases`, and `/actions` rows, and marked the generic row "anything else" — it was pointing `/contents/` readers at the exact anti-pattern the shim blocks.

**P4 (multi-URL denial text): taken.** The closing note gets its own line when several URLs are denied. Single-URL output is byte-identical. Left the `${url}: ` prefix keyed on total URL count — a labelled URL in a 2-URL batch is the useful signal.

**P4 (`assert_suggestion_starts_with`): taken.** Now fails on an empty reason and on an empty prefix.

**Not taken — flagging for a separate change.** Host matching is exact and case-sensitive, so `www.github.com/owner/repo` (a live redirect), `GitHub.com`, `github.com:443`, `user@github.com`, and the trailing-dot FQDN all fall through. Real, and `intercept-github-curl.sh:18` has it too. It's a different fix from "intercept MCP fetch tools" and touches a hook this PR doesn't, so I'd rather not fold it in here. Noted your `tr` over `${host,,}` point — this script is bash 3.2-clean and should stay that way.

PR #276, opened 2026-08-25
testing-handbook-skills: make 15 descriptions routable

kz-tob:
## What

Every one of this plugin's 15 descriptions was a tool-encyclopedia blurb averaging 125 characters — the first half defined the tool, the second half restated it as a trigger:

> `libfuzzer`: Coverage-guided fuzzer built into LLVM for C/C++ projects. Use for fuzzing C/C++ code that can be compiled with Clang.

This is the largest plugin in the repo by skill count and had the thinnest descriptions in it (110–142 chars, against a median in the 300s). Fifteen skills competing on wording like that lose to each other and to siblings elsewhere in the marketplace.

Each description is now three parts: **what it does for the reader**, task first — the `name` field already carries the tool name; **what it covers**, in concrete flags and symbols; then **Use when …**, naming two to four situations in the words a user would type.

> Sets up and runs libFuzzer, the coverage-guided fuzzer built into LLVM, on C/C++ code that compiles with Clang. Covers harness structure, `-fsanitize=fuzzer` builds, corpus and dictionary management, sanitizer integration, and campaign triage. Use when writing or debugging an `LLVMFuzzerTestOneInput` harness, starting fuzzing on a C/C++ library, choosing between libFuzzer and AFL++, or working out why a libFuzzer run finds nothing.

The concrete anchors are the point — `LLVMFuzzerTestOneInput`, `fuzz_target!`, `FuzzedDataProvider`, `afl-clang-fast`, `ASAN_OPTIONS`, `project.yaml`, `llvm-cov`, "an ASan stack trace", "a campaign that finds nothing". Descriptions land at 342–515 chars, well under the 1024 ceiling `validate-skills.py:46` enforces.

**A description may only advertise an anchor its body backs.** Three failed that on the first pass and were re-grounded rather than trimmed: `harness-writing` claimed Ruby it does not contain, `constant-time-testing` named `ctgrind` (absent) instead of Timecop and Valgrind, and `cargo-fuzz` advertised a `cargo fuzz add` subcommand that does not exist. `atheris` advertised `TestOneInput` and `FuzzedDataProvider`, neither of which appeared anywhere in the file — the entry point is renamed to Atheris' own `TestOneInput` spelling, which its docs and error messages use, and `FuzzedDataProvider` now has real coverage.

## Three things found while in there

**The generator would have regressed this immediately.** All four templates under `skills/testing-handbook-generator/templates/` prescribe exactly the shape being removed — `{Summary of what this technique does}. Use when {trigger conditions}.` — and their worked examples were these same descriptions. Fixing only the 15 shipped skills would leave the next generated one thin. Updated all 10 description blocks across the four templates, and added a "Description quality" section to `agent-prompt.md` explaining why the existing rule isn't enough: `MUST include "Use when {trigger}"` is satisfied by "Use for fuzzing C/C++ code," which is how these got written in the first place.

**Nothing rejected a leftover template slot.** `SHORTCODE_PATTERN` needs double braces, so a description shipped with `{language}` still in it passed every check and printed a clean tick. `validate-skills.py` now errors on a `{...}` slot in a description, and `scripts/test_validate_skills.py` holds every description check to a known-bad fixture plus a positive control and a greediness guard across all 15 shipped descriptions. Negative-controlled: neutering `PLACEHOLDER_PATTERN` fails exactly the 2 placeholder cases and nothing else.

**Frontmatter carrying no fields validated clean.** `yaml.safe_load("")` returns `None` and raises nothing, so `extract_frontmatter` returned `(None, None)`; `validate_skill` branches on the error, so a skill with an empty frontmatter block skipped *every* field check — missing name, missing description, trigger phrase, HTML, shortcodes, placeholders — and printed `✓`. A bare scalar took the same path and died on `.get` with an uncaught `AttributeError`. Extraction now pairs both with an error, and `validate_frontmatter` reports rather than returning silently when handed a non-mapping.

**Descriptions moved from folded `>` blocks to one quoted line.** A folded block at this length adds 4–5 lines per file; one quoted line removes 1–2 instead. It's what the rest of the repo uses at this length (`trailmark` at 893 chars, `property-based-testing` at 845), and it's explicitly exempt from the repo validator's plain-scalar rule. Templates updated to match, and `testing.md`'s now-wrong "Use `>` for multi-line" row corrected.

## atheris was split

Giving `FuzzedDataProvider` real coverage took `atheris/SKILL.md` from 519 lines to 552, making a file already over the 500-line limit worse. The structured-input material and the two worked harnesses now live in `structured-input.md` and `examples.md` beside it — the split `agent-prompt.md:229` prescribes for that band. **SKILL.md is 482 lines and passes the line-count check it has failed since it was generated.** The plugin's over-limit list goes from 8 skills to 7; the remaining 7 are untouched by this PR and unchanged from `main`.

## Testing

- All 15 parse under `yaml.safe_load`; each has exactly one trigger phrase, is ≤1024 chars, carries no HTML tags or Hugo shortcodes, and `name` matches its directory.
- Every anchor advertised in a description was grepped against its own body.
- Every `FuzzedDataProvider` method documented is checked against the pybind registration at `src/native/atheris.cc:299-325`, which exposes `remaining_bytes` despite the upstream README omitting it.
- The plugin's own `scripts/validate-skills.py`: **zero description errors**, `atheris` now passing. The 7 remaining line-count failures are pre-existing and identical to `main`.
- `make validate` clean — 0 errors. `make python-tests`: 16 passed for this plugin. `make lint` and `prek` clean.

Version `1.1.1` → `1.2.0`, MINOR. 28 files, +422/−154.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -195,7 +195,7 @@
     },
     {
       "name": "testing-handbook-skills",
-      "version": "1.1.2",
+      "version": "1.2.0",
       "description": "Skills from the Trail of Bits Application Security Testing Handbook (appsec.guide)",
       "author": {
         "name": "Paweł Płatek"
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -195,7 +195,7 @@
     },
     {
       "name": "testing-handbook-skills",
-      "version": "1.1.1",
+      "version": "1.1.2",
       "description": "Skills from the Trail of Bits Application Security Testing Handbook (appsec.guide)",
       "author": {
         "name": "Paweł Płatek"
--- a/plugins/testing-handbook-skills/.claude-plugin/plugin.json
+++ b/plugins/testing-handbook-skills/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "testing-handbook-skills",
-  "version": "1.1.1",
+  "version": "1.1.2",
   "description": "Skills from the Trail of Bits Application Security Testing Handbook (appsec.guide)",
   "author": {
     "name": "Paweł Płatek"
--- a/plugins/testing-handbook-skills/.claude-plugin/plugin.json
+++ b/plugins/testing-handbook-skills/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "testing-handbook-skills",
-  "version": "1.1.2",
+  "version": "1.2.0",
   "description": "Skills from the Trail of Bits Application Security Testing Handbook (appsec.guide)",
   "author": {
     "name": "Paweł Płatek"
--- a/plugins/testing-handbook-skills/README.md
+++ b/plugins/testing-handbook-skills/README.md
@@ -113,7 +113,7 @@ Each generated skill:
 
 ## Skills Cross-Reference
 
-This graph shows the 16 generated skills and their cross-references (from the Related Skills section of each skill). Only links between actually generated skills are shown.
+This graph shows the 14 generated skills and their cross-references (from the Related Skills section of each skill). Only links between actually generated skills are shown.
 
 ```mermaid
 graph TB
@@ -218,7 +218,6 @@ graph TB
 |------|--------|
 | Fuzzers (6) | libfuzzer, aflpp, libafl, cargo-fuzz, atheris, ruzzy |
 | Techniques (6) | harness-writing, address-sanitizer, coverage-analysis, fuzzing-dictionary, fuzzing-obstacles, ossfuzz |
-| Tools (2) | semgrep, codeql |
 | Domain (2) | wycheproof, constant-time-testing |
 
 **Note:** Some skills reference planned/external skills not yet generated (e.g., `honggfuzz`, `fuzzing-corpus`, `sarif-parsing`). Run `validate-skills.py` to see the full list.
--- a/plugins/testing-handbook-skills/README.md
+++ b/plugins/testing-handbook-skills/README.md
@@ -135,11 +135,6 @@ graph TB
         ossfuzz[ossfuzz]
     end
 
-    subgraph Tools
-        semgrep[semgrep]
-        codeql[codeql]
-    end
-
     subgraph Domain
         wycheproof[wycheproof]
         constant-time-testing[constant-time-testing]
@@ -170,10 +165,6 @@ graph TB
     libafl -.-> cargo-fuzz
     ruzzy -.-> libfuzzer
     ruzzy -.-> aflpp
-
-    %% Tool ↔ Tool alternatives
-    semgrep -.-> codeql
-    codeql -.-> semgrep
 
     %% Technique → Fuzzer references
     harness-writing --> libfuzzer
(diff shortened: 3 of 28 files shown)


kz-tob:
All taken. Pushed 7db16ab.

**P2 + the two related P3s were the important ones, and all three were mine.** Verified each before fixing:

| Claim | Check | Result |
|---|---|---|
| `harness-writing` has no Ruby | `rg -in 'ruby\|ruzzy\|gem\b\|\.rb\b'` | 0 hits outside line 4 |
| Python only as a Related Skills row | `rg -in 'python\|atheris'` | 1 hit, line 581 |
| `ctgrind` absent | `rg -irn ctgrind` across the plugin | 1 hit — my own description |
| `cargo fuzz add` absent | `rg -o 'cargo (\+nightly )?fuzz [a-z-]+'` | init, run, coverage, crash, --version |

Fixes, using only anchors I re-grepped as present:

- **`harness-writing`** → "for C/C++ and Rust", with `LLVMFuzzerTestOneInput` (12 sites) and `fuzz_target!` (16) as the triggers. You were right that this would have taken "harness for my Ruby gem" from `ruzzy` and handed back a file with no Ruby in it. Synced the template's worked example at `technique-skill.md:267` too.
- **`constant-time-testing`** → `ctgrind` out, **Timecop** (20 hits) and **Valgrind** (10) in. Also took your last P3 in the same edit: it now leads with *measuring a running implementation* and closes with "Not for statically inspecting compiler output — the constant-time-analysis plugin covers that", restoring the split that sibling documents at its own `SKILL.md:21`. Good catch that I'd applied my own "add a closing Not for" rule to one skill and not this one.
- **`cargo-fuzz`** → dropped `add`; added the nightly requirement (25 hits) and `cargo fuzz coverage` (17), which as you noted the skill does cover and the description had omitted.

**P4 (nothing rejects leftover placeholders): taken, and this is the one I'd have most regretted leaving.** `validate-skills.py` now errors on a description containing a `{...}` slot, and `test_validate_skills.py` holds every description check to a known-bad fixture plus a positive control and a greediness guard over all 15 shipped descriptions. Negative-controlled: neutering `PLACEHOLDER_PATTERN` fails exactly the 2 placeholder cases and nothing else.

One thing that surfaced writing it — the suite is stdlib-only. It would have been the only one in the repo whose module-under-test imports a third-party package, and CI runs `uv run --no-project --with pytest`, which has no pyyaml. My first version passed locally only because of a `.venv` I'd left in the directory. Confirmed the fix by deleting it: `import yaml` now raises `ModuleNotFoundError` in that environment and all 12 tests still pass.

**P3 (`agent-prompt.md:76` pointer): taken** — my section is under `## Critical Rules`, not Pass 2. Now reads `(see "Description quality" under Critical Rules)`.

**P3 (`testing.md` still encodes the two-part bar): taken, and you're right this was the likeliest route back.** Lines 127, 349, and 449 updated; the checklist item now also requires that every anchor a description advertises appears in the body — which is exactly the class of bug you found above.

**P4 (version undersized): taken.** `1.2.0`.

**P4 (README inventory stale): taken** — "16 generated skills" → 14, and dropped the `Tools (2) | semgrep, codeql` row for skills that live in `static-analysis`.

**Not taken:** the `yq` inconsistency at `testing.md:108/112/136` vs 105. Real and worth fixing, but it's a documented shell recipe rather than executed code, I can't run `yq` here to establish current behaviour either, and it's unrelated to descriptions. Left for its own change. Root `README.md:69` still says "static analysis" for this plugin, same root cause as the table row — also separate.

Re-verified after all changes: plugin validator reports zero description errors (the 8 line-count failures are the pre-existing ones, unchanged), `make validate` 0 errors, `make python-tests` 12 passed, `make lint` and `prek` clean.

PR #278, opened 2026-08-25
Give differential-review a trigger, and name every component in its README

kz-tob:
## What

**`differential-review`'s description listed capabilities and never named a situation:**

> Performs security-focused differential review of code changes (PRs, commits, diffs). Adapts analysis depth to codebase size, uses git history for context, calculates blast radius, checks test coverage, and generates comprehensive markdown reports. Automatically detects and prevents security regressions.

It now keeps the capabilities and closes with the triggers the plugin README already documented in its own "When to Use" section — *reviewing a PR, commit, or diff for security vulnerabilities; checking whether a change re-introduces a previously fixed bug; asking what else a change could break; finding which modified code has no test covering it.* Also moved to the single-line quoted form the repo validator exempts from the plain-scalar rule. `allowed-tools` unchanged.

**The same plugin's README never mentioned `adversarial-modeler`,** which is what Phase 5 dispatches for HIGH RISK changes. Checking whether that was isolated turned up more of it:

| Plugin | Gap | Version |
|---|---|---|
| `differential-review` | Trigger clause; Components table naming the skill, the `/differential-review:diff-review` command with its argument shape, and `adversarial-modeler` | 1.1.1 → 1.1.2 |
| `zeroize-audit` | Agent table listed 8 of 11. Missing `0-preflight` — which gates the whole run, "if any critical check fails, the run stops here" — plus `5b-poc-validator` and `5c-poc-verifier` | 0.3.0 → 0.3.1 |
| `constant-time-analysis` | README documents the `ct-analyzer` CLI end to end and never says the plugin also ships a skill and a command | 0.2.2 → 0.2.3 |
| `entry-point-analyzer` | Lists natural-language phrases that trigger the skill, but never names the skill or its `/entry-point-analyzer:entry-points` command | 1.0.1 → 1.0.2 |
| `claude-in-chrome-troubleshooting` | Names `chrome-mcp-troubleshooting` — not guessable from the plugin name | 1.1.1 → 1.1.2 |
| `culture-index` | Names `interpreting-culture-index` — likewise | 1.1.2 → 1.1.3 |
| `devcontainer-setup` | Names the skill | 0.2.1 → 0.2.2 |

The three `zeroize-audit` agents all appear in the phase diagram immediately above the table, which is why they read as present — a reader consulting the table for what runs when gets an incomplete pipeline and no sign of it.

All seven bumps are PATCH.

## Testing

The drift scan covers skills, agents, **and** commands across all 42 plugins — **119 components checked, zero gaps**. My first pass only checked skills and agents, and adding commands is what surfaced `entry-point-analyzer`, so the scan is only as good as its coverage. It has a floor (`checked < 100` exits non-zero), because a scan that silently inspects nothing prints "✓ all clean" exactly like one that inspects everything.

Also asserted that all 42 `plugin.json` / `marketplace.json` **version and description** pairs still agree, since seven bumps landed at once — the marketplace entries put `description` between `name` and `version`, which broke my first bump script and left it half-applied, so this is verified rather than assumed.

`make validate` clean (0 errors, only the pre-existing `testing-handbook-skills` length warnings), `prek` passes.

Rebased onto `8c1b267` so `constant-time-analysis` bumps from the 0.2.2 that #277 shipped.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -29,7 +29,7 @@
     },
     {
       "name": "burpsuite-project-parser",
-      "version": "1.0.1",
+      "version": "1.0.2",
       "description": "Search and extract data from Burp Suite project files (.burp) for security analysis",
       "author": {
         "name": "Will Vandevanter"
@@ -147,7 +147,7 @@
     },
     {
       "name": "semgrep-rule-creator",
-      "version": "1.2.3",
+      "version": "1.2.4",
       "description": "Create custom Semgrep rules for detecting bug patterns and security vulnerabilities",
       "author": {
         "name": "Maciej Domanski"
@@ -177,7 +177,7 @@
     },
     {
       "name": "static-analysis",
-      "version": "1.4.0",
+      "version": "1.4.1",
       "description": "Static analysis toolkit with CodeQL, Semgrep, and SARIF parsing for security vulnerability detection",
       "author": {
         "name": "Axel Mierczuk & Paweł Płatek"
@@ -283,7 +283,7 @@
     },
     {
       "name": "git-cleanup",
-      "version": "2.3.0",
+      "version": "2.3.1",
       "description": "Safely analyzes and cleans up local git branches and worktrees by categorizing them as merged, squash-merged, superseded, or active work.",
       "author": {
         "name": "Henrik Brodin",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -126,7 +126,7 @@
     },
     {
       "name": "mutation-testing",
-      "version": "1.0.1",
+      "version": "1.0.2",
       "description": "Configures mewt or muton mutation testing campaigns — scopes targets, tunes timeouts, and optimizes long-running runs. Use when the user mentions mewt, muton, mutation testing, or wants to configure or optimize a mutation testing campaign.",
       "author": {
         "name": "Trail of Bits",
@@ -204,7 +204,7 @@
     },
     {
       "name": "trailmark",
-      "version": "0.11.1",
+      "version": "0.11.2",
       "description": "Builds multi-language source and binary code graphs for security analysis: call graphs, attack surface mapping, blast radius, taint propagation, complexity hotspots, entry point enumeration, proxy/unresolved-call tracking, type/reference analysis, and structural diffs. Creates bounded, graph-informed source packets for delegating focused work to constrained subagents. Generates Mermaid diagrams, runs graph-informed mutation testing triage (genotoxic), generates mutation-driven test vectors (vector-forge), extracts crypto protocol message flows, converts Mermaid diagrams to ProVerif models, projects SARIF/weAudit/binary findings onto code graphs, triages single findings with graph evidence, gates branch diffs for structural review regressions, and expands seed findings into variant-neighborhood candidates. Use when analyzing call paths, slicing source context for smaller models, mapping attack surface, visualizing code architecture, triaging survived mutants, generating cryptographic test vectors, diagramming crypto protocols, formally verifying protocols, augmenting audits with static analysis findings, deciding whether one candidate issue is reachable, reviewing graph-level PR risk, or seeding variant analysis.",
       "author": {
         "name": "Scott Arciszewski",
@@ -223,7 +223,7 @@
     },
     {
       "name": "c-review",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "description": "Comprehensive C/C++ security code review, with coverage verified against a parse of the source",
       "author": {
         "name": "Paweł Płatek"
@@ -252,7 +252,7 @@
     },
     {
       "name": "insecure-defaults",
-      "version": "2.0.0",
+      "version": "2.0.1",
       "description": "Detects insecure default configurations including hardcoded credentials, fallback secrets, weak authentication defaults, and dangerous values in production",
       "author": {
         "name": "Trail of Bits",
@@ -323,7 +323,7 @@
     },
     {
       "name": "let-fate-decide",
-      "version": "1.2.2",
+      "version": "1.2.3",
       "description": "Draws the 12 Houses of the Zodiac Tarot spread using cryptographic randomness to add 100+ bits of entropy to vague or underspecified planning. Interprets the spread to guide next steps. Use when feeling lucky, invoking heart-of-the-cards energy, or when prompts are ambiguous.",
       "author": {
         "name": "Scott Arciszewski",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -38,7 +38,7 @@
     },
     {
       "name": "claude-in-chrome-troubleshooting",
-      "version": "1.1.1",
+      "version": "1.1.2",
       "description": "Diagnose and fix Claude in Chrome MCP extension connectivity issues",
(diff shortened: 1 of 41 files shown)


kz-tob:
All taken, including the P4 you argued hardest for. Pushed 07a616b.

**P2 (PoC rows one phase early): confirmed and fixed.** My rows said Phase 4; the diagram three lines above them, `SKILL.md:179-180`, and `workflows/phase-5-poc-validation.md` all say Phase 5, steps 5a/5b. You're right that "Wave 5a" is a label that exists nowhere. The failure you describe is the real cost — the table is the artifact this PR designates as *what runs when*, and it would have sent a debugger to `phase-4-poc-generation.md`, which spawns only `5-poc-generator`.

**P3 (count above the table): fixed** — `10 agents across 8 phases` → `11 agents across 9 phases (0-8)`. You were right on both halves; `SKILL.md:167` already said 11.

**P4 (Phase 0 diagram credits the orchestrator): fixed** in the same pass — it now credits `0-preflight`, matching `SKILL.md:188` and the row I added. Fair hit that diagram and table disagreed inside the eight lines I touched.

**P3 (advertises a dispatch that never happens): fixed.** Verified `subagent_type` appears nowhere under the plugin and the only instruction is the prose at `SKILL.md:77`. Namespaced both that line and the Name column to `differential-review:adversarial-modeler`, and softened "Dispatched" to "The skill delegates to it". The inconsistency you spotted — namespaced skill and command rows, bare agent row, the one form that fails — was the tell.

**P3 (`static-analysis` gap survives the sweep): fixed, and it was the most useful finding here.** `codeql-build` documented. But your framing was the valuable part, so I acted on it rather than just patching the instance — see below.

**P4 (the sweep is not repeatable): taken.** Doing this by hand three times found eight gaps and missed two more of the identical shape. It's now a validator check: a README must name every skill, agent, command, and workflow its plugin ships, reading `meta.name` for workflows exactly as you specified. It refuses a run that inspected zero components and prints the denominator (`129 components named in their README`), and six self-test assertions hold it to known-bad fixtures — including the filename-vs-`meta.name` case specifically, since that's the one a glob cannot see. Self-test floor 82 → 88.

**It found a ninth gap on its first run**, exactly the shape you predicted: `git-cleanup` ships as `/git-cleanup:git-cleanup-analysis` but its README cites `workflows/analyze-branches.js` four times and that name never. My earlier claim that the other nine workflows were all named was wrong — I'd checked filenames.

Negative-controlled both directions: neutering the finding fails the self-test at "skill absent from README"; neutering discovery trips the zero-component guard in `make validate`.

**P4 (stale `trailofbits:` name keys): taken.** Removed from all five, not just the three I documented — the newest command files carry no `name:` at all, so it's a dead key contradicting the docs I'm adding.

**P4 (line counts): fixed** — measured rather than estimated. `reporting.md` was the worst at 369 vs "~120".

**P3 (`{baseDir}` in constant-time-analysis): confirmed real, not fixed here.** `ct_analyzer/` sits at the plugin root while `SKILL.md:46` invokes `{baseDir}/ct_analyzer/analyzer.py`, and every other plugin resolves `{baseDir}` to the skill directory — zeroize-audit's `{baseDir}/tools/`, culture-index's `{baseDir}/scripts/`, and this same file's relative `references/` links. At most one convention is right and I can't settle which from here. It needs a decision — move the package under `skills/`, or rewrite the paths — that changes how the analyzer is packaged, so it belongs in its own PR. I did drop the "both of which drive that same tool" phrasing my row rested it on.

Fixture note: adding the check broke 13 self-test fixtures that assert zero errors while adding an unnamed component. Rather than weaken the check, `_build_demo`'s README now names the fixture component vocabulary, so each fixture stays isolated to the checker it targets — a plugin shipping an unnamed component genuinely is invalid, so the fixtures were the thing that was wrong.

`make validate`, `make self-test` (88), `make lint`, and `prek` all clean.

kz-tob:
Both P2s fixed. Pushed 7dd2edb.

**P2 — the gate was a substring test.** You're right, and the framing was the useful part: `name not in text` reads as thorough and could not fail for a large share of what it counted. Rather than special-case the four, I matched by kind, following your proposal:

- **commands and workflows** — reachable only as `/<plugin>:<name>`, so that literal is now required. It is the only string a user can type; nothing else helps them.
- **agents** — dispatched by identifier, never typed as prose, so an identifier-shaped mention is required (backticked, an `agents/<name>` path, or namespaced). This is the `draw` case: my first tightening still passed it on "(draw cards instead)", which is the same bug one word narrower.
- **skills** — genuinely referred to by bare name in prose and tables, so only a delimited occurrence. That is what stops "draws" counting as `draw` and `semgrep-rule-creator` as `semgrep-rule`.

**It surfaced seven gaps, not four** — your four plus `insecure-defaults:audit-pipeline`, `mutation-testing`'s skill, and `trailmark`'s `code-slice-worker` (its tree lists `code-slice-worker.md`, never the dispatchable name). All seven fixed.

One correction to my own first pass: I initially put `/` in the skill/agent lookbehind, which failed 15 plugins by also rejecting legitimate `skills/<name>/SKILL.md` path mentions. Removed — commands and workflows no longer need it, since they require the slash form outright.

**P2 — `adversarial-modeler` still bare.** Confirmed and fixed. Line 77 was the decision-tree mention; `SKILL.md:96` is the `## Agents` block carrying "Delegate to this agent", which is the one a model acts on — so the runtime failure my previous commit claimed to fix did survive. It now carries the namespaced name and states that a bare `adversarial-modeler` is unregistered.

Also taken:

- **P3 (total floor hides the likeliest breakage)** — now per-kind. Your `skill_files()` scenario was exact: 81 of 129, and the total would have stayed at 48 and printed a tick.
- **P4 (`workflow_names` fragility)** — anchored to `export const meta`, so an earlier `name:` in a comment cannot win; `.mjs` added; a file with no meta block is a helper, not a workflow; a templated name is skipped rather than falling through to the stem the docstring calls wrong.
- **P4 (guard had no fixture)** — three added, including the comment-before-meta case.
- **P3 (AGENTS.md)** — the new hard failure is documented in "What the validator enforces", with the per-kind rules.

Self-test 88 → 96, floor raised to match. Negative-controlled both new rules: neutering the command/workflow branch fails at "command satisfied only by a same-named script path", neutering the agent branch at "agent satisfied only by prose".

**P4 (zeroize step labels): you're right and my commit message was wrong.** `SKILL.md:198-201` says "Step 1"/"Step 2", not 5a/5b — I checked the workflow file and the diagram and wrote the message as though SKILL.md agreed. The README now matches `phase-5-poc-validation.md`; SKILL.md's local numbering is a separate cosmetic inconsistency I have not touched.

**P3 (`name:` vs #52's claim): recording how I determined it, as you asked.** Three pieces of evidence, none of them a CLI run: the published command frontmatter fields do not include `name`; the three newest command files in this repo carry no `name:` at all and their commands work; and #49 created these as `commands/trailofbits:burp-search.md`, so the namespace came from the filename, which is what #52's rename actually changed. That is consistent with the key being inert, but I could not execute the CLI to settle it — so if you want this held until someone can, say so and I will revert that hunk. It is five lines.

**P3 (`{baseDir}`)** — still deferred, for the reason given last round.

Five more PATCH bumps for the newly-touched plugins (c-review, insecure-defaults, let-fate-decide, mutation-testing, trailmark), 16 total. Verified every touched plugin is bumped and all 42 version+description pairs agree. `make validate`, `make self-test` (96), `make lint`, `prek` clean.

PR #281, opened 2026-08-26
static-analysis: detect every language run-scans.sh can scan

kz-tob:
## What

Step 1 of the Semgrep scan workflow globbed for 14 file extensions. `includes_for` in `scripts/run-scans.sh` carries globs for 41.

Detection is the only input to ruleset selection, so a category Step 1 never reports is one Step 2 never selects and the scan never runs. These rulesets existed in `references/rulesets.md` and could not be reached:

| Unreachable | Ruleset |
|---|---|
| `.yaml`, `.yml` | `p/yaml`, `p/kubernetes`, `p/github-actions`, `p/cloudformation` |
| `.cs` | `p/csharp` (170+ rules) |
| `.kt`, `.kts` | `p/kotlin` |
| `.scala` | `p/scala` |
| `.swift` | `p/swift` |
| `.ex`, `.exs` | `p/elixir` |
| `.cls`, `.trigger` | `p/apex` |
| `.sol` | Decurity third-party rules |
| `.json` | `r/json.aws`, and JSON-format `p/cloudformation` |

A header-only C++ tree (`.h`/`.hh`/`.hpp`/`.hxx`), a `.cc`-only one, and an ESM-only `.mjs` package matched nothing either. In every case the scan completed and the report read clean rather than incomplete — the failure mode `AGENTS.md` calls the most expensive class of bug in this repo.

Separately, the framework markers were written without the `**/` prefix that every language pattern had, so they matched only the target root. A monorepo keeping `package.json` in `packages/*/` or `go.mod` in `services/*/` got no framework rulesets at all.

## Changes

**`skills/semgrep/workflows/scan-workflow.md`**
- Detection list is now the full union of the `includes_for` globs, with the invariant stated so the two stay in sync. `.mts`, `.cts`, `.C`, `Containerfile` and `Dockerfile.prod` are absent from both, because semgrep does not parse them.
- Category table gained the rows the new patterns resolve to, so no glob dangles without a destination. Every category name is one `canonical_lang` in `run-scans.sh` folds (`C#`→`csharp`, `C/C++`→`cpp`, `GitHub Actions`→`github-actions`).
- **`.yaml`/`.yml` and `.json` are assigned by content, not by extension.** Both are near-universal in repositories with no infrastructure to scan, so keying a category off the extension would attach an AWS IAM ruleset to every scan. Keying off nothing would leave those rulesets unreachable — and an unselected category never enters `rulesets.json`, so it cannot show up in `coveredNothing`, `failed` or `skipped` either, which is the clean-looking incomplete report this PR exists to remove. Both are globbed, then sampled and assigned on markers: `apiVersion:`+`kind:` → Kubernetes, `AWSTemplateFormatVersion` → CloudFormation, a `Statement` array with `Effect` members → JSON, and for JSON specifically there is no catch-all — `package.json`, `tsconfig.json`, lockfiles and editor settings yield **no category**.
- Framework markers prefixed with `**/`, and pick up `composer.json` and `requirements.txt`, which `rulesets.md` keys on but the list omitted.

**`workflows/semgrep-scan.js`** — the detect-phase prompt enumerated the same 13 categories, so the ungated path had the identical blind spot. `SKILL.md` advertises it as the same scan without the approval gate, so it gets the same category list, extension guidance, and YAML/JSON content rules.

Version 1.4.1 → 1.4.2.

## Verification

- Diffed the doc list against `includes_for` in both directions: 41 = 41, no drift either way.
- Checked that every `includes_for` case arm is named in the JS detect prompt.
- `make validate` clean; reference count 398 → 399, so the new `run-scans.sh` link resolves. Version-increment check passes against `origin/main`.
- `tests/run_workflow_tests.sh` (61 + 62 assertions), `tests/run_scan_tests.sh` (78), `make js-tests` (26/26), `make lint` clean.

Not run locally, none related to this change: `make bats` and `make eval-self-tests` need `bats` and `timeout`, neither installed on this machine; `make python-tests` has 31 pre-existing failures in `constant-time-analysis` (missing cross-compilation toolchains) and `dwarf-expert`. `static-analysis`'s own Python suites pass.

## History

`822f5da` dropped `.json` from detection outright. Review pointed out that made `r/json.aws` and JSON CloudFormation unreachable with no way for a user to ask for them, so `043b7dd` replaced the removal with content-based assignment, which keeps the false-positive fix and restores reachability.

## Follow-up

Nothing enforces the doc-list ↔ `run-scans.sh` invariant, which is why it drifted; the extension list now lives in three places. A test extracting the `includes_for` case arms and the backticked globs from `scan-workflow.md`, asserting set equality and failing when either extraction yields zero patterns, would fit in `plugins/static-analysis/tests/`.

Review P3/P4s left open: `c` vs `cpp` sharing one extension group in the JS prompt, unbounded YAML/JSON sampling leaving a missed category untraceable, `**/` framework markers reading vendored manifests, Pro-only categories detected on OSS runs, `.cls` overlapping LaTeX, and the Glob call count.


kz-tob:
P2 fixed in 043b7dd. The finding was right, including the part I should have caught myself: I wrote an escape hatch without checking it was reachable.

## P2 · `scan-workflow.md:86` — dropped JSON, escape hatch has no input path

Confirmed. `parseArgs` (`:37`) takes only `target`, `out`, `mode`, `jobs`, `skill`; Select sees only what Detect reported; and the Step 3 plan lists only detected categories, so the gated path prompts nobody either. "Name the `json` or `cloudformation` category explicitly instead" described a capability that does not exist. And your point that it is *strictly worse* than before `822f5da` lands — an unselected category never enters `rulesets.json`, so it cannot surface in `coveredNothing`, `failed` or `skipped`, which is exactly the clean-looking incomplete report this PR was opened to remove.

I took the narrower fix you proposed: key on content, as the YAML block already does. `.json` is globbed again, but the category comes from markers rather than the extension —

- `"AWSTemplateFormatVersion"`, or `"Resources"` with a `"Type": "AWS::…"` member → `cloudformation`
- a `"Statement"` array whose elements have `"Effect"` → `json`, the IAM shape `r/json.aws` targets
- anything else — `package.json`, `tsconfig.json`, lockfiles, editor settings → **no category**

That keeps the reason `.json` was dropped (no IAM ruleset on every scan) while restoring reachability. Sampling is steered at paths suggesting infrastructure — `iam/`, `policies/`, `cloudformation/`, `infra/`, `*template*.json` — since build config outnumbers policies and an unordered sample will miss them; that is a partial mitigation of your separate P3 about unbounded sampling, not a fix for it.

`semgrep-scan.js:253-256` had the identical gap and got the same treatment.

Side effect worth noting: describing both `.yaml` and `.json` as content-assigned in one place removes the "union minus one exclusion" wording, so the detection list is the full union again. Verified 41 = 41, no drift in either direction, and separately that every `includes_for` case arm is named in the JS prompt.

## Scope

I drafted the `c` vs `cpp` P3 fix into the same block — three lines, and your point that `includes_for c` is `*.c *.h` so choosing `c` for a C++ tree scans zero files is correct — then **removed it before committing.** It is not made reachable by the P2 fix and the request was for the P2, so it should be its own change rather than riding along unreviewed.

I did update the PR description, which was your P4: it described pre-`822f5da` behaviour and would have become the merge commit message claiming `r/json.aws` became reachable when it had become unreachable. It now covers the content-assignment design, carries a History section explaining the `822f5da` → `043b7dd` sequence, and states the verification as 41 = 41 rather than the earlier "zero drift".

## Still open

Your remaining findings, unaddressed and worth their own passes:

- **P3** `c`/`cpp` sharing one extension group — drafted, deliberately not included
- **P3** unbounded YAML/JSON sampling: a missed category is untraceable. Your observation that `Grep` is absent from the skill's `allowed-tools` while the JS agent has it is the useful detail here
- **P3** `**/` framework markers reading `node_modules`/`vendor` manifests
- **P3** the invariant now restated in a third place with nothing enforcing it — this is the Follow-up test, and the drift it would have caught is the reason this PR exists
- **P4** Pro-only categories (apex, elixir, csharp) detected on OSS runs; `.cls` overlapping LaTeX/VB6; 48 Glob calls with counts that are floors once truncated
- **P4** PATCH vs MINOR for a change taking detection from 14 extensions to 40. Fair; I would take MINOR if you want it, since the version is the one thing clients act on

kz-tob:
P3 `c`/`cpp` fixed in b7f8612 — this is the change I drafted and pulled from 043b7dd for being out of scope.

## The finding

Confirmed against `run-scans.sh:53-54`:

```sh
c)   echo '*.c *.h' ;;
cpp) echo '*.c *.cc *.cpp *.cxx *.h *.hh *.hpp *.hxx' ;;
```

`cpp` is a strict superset. Reporting `c` for a `.cpp`/`.hpp` tree scans with `--include=*.c --include=*.h`, opens zero files, exits 0, and lands in `coveredNothing` — so the C/C++ rules never read the source, and the only trace is a category that covered nothing, which reads like an empty directory rather than a mis-scoped scan.

## Fix

The detect prompt now states the rule as "report the SUPERSET, never the narrower one", with the `includes_for` sets and the zero-file consequence spelled out so the reason travels with the instruction.

I covered `javascript`/`typescript` in the same block rather than as a second special case: `javascript` already carries the `.ts`/`.tsx` globs — `run-scans.sh:45` says so at the `includes_for` comment — so it is the same superset relation, and reporting both scans the same files twice under two rulesets. One rule stated once is less likely to rot than two adjacent exceptions.

## The gated path needs no change

You noted `scan-workflow.md:101` folds all eight extensions into one `C/C++` row. I checked rather than assumed: `canonical_lang` (`run-scans.sh:82`) maps `'c/c++' | 'c++' | cxx` to `cpp`, so the markdown table already resolves to the superset. Same for its single `JavaScript/TypeScript` row, which folds to `javascript`. So the asymmetry you flagged in "the same category list" is real, but it runs in the safe direction — the folded table was never the buggy path, and the fix belongs only in the JS prompt.

## Note on the branch

Your `main` merge (0f7ae1d) landed while I was working, so I rebased this commit on top of it rather than force-pushing over it. `#282` and `#283` are now in main and pulled in through that merge. Re-ran everything post-rebase: `run_workflow_tests.sh` 61+62, `run_scan_tests.sh` 78, `make js-tests` 26/26, `make validate` clean, version still 1.4.2 against main's 1.4.1.

## Still open

- **P3** unbounded YAML/JSON sampling — a missed category is untraceable. Your point that `Grep` is absent from the skill's `allowed-tools` while the JS agent has it is the constraint that shapes any fix.
- **P3** `**/` framework markers reading `node_modules`/`vendor` manifests.
- **P3** the invariant restated in a third place with nothing enforcing it — the Follow-up test, and the highest-value one left.
- **P4** Pro-only categories detected on OSS runs; `.cls` overlapping LaTeX/VB6; 48 Glob calls with truncated counts as floors; PATCH vs MINOR.

PR #283, opened 2026-08-27
constant-time-analysis: pair ilspycmd's TFM with a matching runtime

kz-tob:
## What

The tool-store fallback in `_get_il_output` globbed for `*/ilspycmd/*/tools/net8.0/any/ilspycmd.dll`.

ilspycmd 9.x installs under `tools/net9.0/`, so on a newer install the glob matched nothing, the loop body never ran, and C# IL analysis fell through to `monodis` and then to `"IL disassembly tools not available. Install ILSpy CLI: dotnet tool install -g ilspycmd"` — while that very command produces the layout the glob could not see.

## Why the obvious fix is wrong

Widening the glob to `net*` alone would have introduced a different bug. The next lines run the assembly under a Homebrew **`dotnet@8`** runtime, and that pairing is deliberate: ilspycmd installs framework-dependent, and .NET does not roll forward across a major version by default, so a `net9.0` assembly will not start on an 8.0 runtime. A wide glob with a pinned runtime converts a silent no-match into a silent failed exec.

The TFM found on disk has to decide the runtime; the two are not independent choices.

## Change

`_tfm_version` parses the moniker; `_il_via_versioned_runtime` holds the fallback so `_get_il_output` stays flat.

- The discovered TFM's major version selects `dotnet@<major>`.
- Candidates are tried newest-first, sorting the parsed `(major, minor)` tuple rather than the moniker string — lexically `"net10.0"` sorts below `"net8.0"`.
- A TFM whose runtime is absent falls back to an older TFM that has one. The arbitrary `break  # Only try the first matching dll` is gone.
- Store entries that name no runtime major (`netstandard2.0`, `net48`) are skipped rather than producing a `dotnet@netstandard2` path.

Version 0.2.3 → 0.2.4.

## Verification

`TestCSharpILRuntimePairing` (6 tests) uses a fake tool store and fake runtime paths, so it needs no dotnet install — the absence of one is why this went unnoticed.

The suite was mutation-tested rather than trusted green:

| Source under test | Result |
|---|---|
| Fix as written | 6/6 pass |
| Old pinned `net8.0` glob | **2 fail**, incl. "net9.0 store assembly should have been found" |
| Naive widening (wide glob, runtime pinned to `dotnet@8`) | **4 fail**, incl. "should not pair a net9.0 assembly with dotnet@8" |

So the tests catch both the original bug and the wrong fix.

`ruff check` and `format --check` clean. `make validate --base-ref origin/main` clean, version-increment passes.

Full `ct_analyzer` suite: 30 failed / 96 passed / 14 skipped, against a baseline of 30 failed / 90 passed / 14 skipped — same 30 failures, +6 from this PR. The three distinct failures are pre-existing and environmental on my machine (riscv64 LLVM target, Rust cross `std`, Java tooling); none are in the C# path.

## Follow-up, deliberately out of scope

Setting `DOTNET_ROLL_FORWARD=Major` and using the system dotnet would drop the Homebrew dependency and make this fallback work on Linux, where no `dotnet@N` formula exists. That is new capability rather than a fix to the glob, so it belongs in its own change.

Related: #281, #282 (same glob-audit pass, different plugins). Independent branches, any order.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -47,7 +47,7 @@
     },
     {
       "name": "constant-time-analysis",
-      "version": "0.2.3",
+      "version": "0.2.4",
       "description": "Detect compiler-induced timing side-channels in cryptographic code",
       "author": {
         "name": "Scott Arciszewski",
--- a/plugins/constant-time-analysis/.claude-plugin/plugin.json
+++ b/plugins/constant-time-analysis/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "constant-time-analysis",
-  "version": "0.2.3",
+  "version": "0.2.4",
   "description": "Detect compiler-induced timing side-channels in cryptographic code",
   "author": {
     "name": "Scott Arciszewski",
--- a/plugins/constant-time-analysis/ct_analyzer/script_analyzers.py
+++ b/plugins/constant-time-analysis/ct_analyzer/script_analyzers.py
@@ -11,6 +11,7 @@ that work at the bytecode/opcode level rather than native assembly.
 
 import os
 import re
+import shutil
 import subprocess
 import sys
 import tempfile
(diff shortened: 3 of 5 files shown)


kz-tob:
Both P2s addressed in 05cb3b5. The P3 `OSError` widening came along with them for a reason given below; the remaining P3/P4s are untouched.

## P2 · `script_analyzers.py:3069` — depends on a `dotnet@9` keg existing

**The premise is wrong, and I checked before acting on it.** You asked me to run `brew info dotnet@9`:

```
==> dotnet@9: stable 9.0.120 (bottled) [keg-only]
$ brew search '/^dotnet(@[0-9]+)?$/'
dotnet    dotnet@6    dotnet@8    dotnet@9
```

The keg exists, so the ilspycmd 9.x + system dotnet 10 case the PR was written for does work as shipped. .NET 9 being STS did not stop homebrew-core carrying it.

The general concern holds regardless, so I fixed it anyway: `dotnet@7` *was* dropped at EOL, `dotnet@10` is not there yet, and no Linux distribution ships a versioned keg at all. An exact-major keg is now preferred rather than required, with the unversioned Homebrew prefixes, the official macOS/Linux installer paths, and the PATH dotnet tried after it under `DOTNET_ROLL_FORWARD=Major`.

That makes your deferred-follow-up point half right: roll-forward is now in scope, but as the *general* path rather than the fix for a keg that turned out to exist. It is set only on the non-exact candidates — putting it on the exact-major keg would mask a genuinely mismatched pair instead of letting it fail, which is the property `test_runtime_major_must_match_the_assembly` pins. I took your "cheaper partial" suggestion for the paths and added `/usr/share/dotnet/dotnet` for Linux.

The PATH candidate resolves through `shutil.which`, since `dotnet_path` defaults to the bare name `"dotnet"` and `Path("dotnet").exists()` tests the cwd.

## P2 · `references/vm-compiled.md:306` — doc says the wrong runtime

Correct, and it was the only user-facing remedy for that error. Rewritten: it now tells the reader to read the TFM out of the tool store rather than assuming .NET 8, names the keg gap explicitly, and explains that the roll-forward path means no install is strictly required. The store lookup uses `find`, not `ls` on a glob — the pitfall #282 documents.

## P3 · `except FileNotFoundError` too narrow — included

Not scope creep: this change makes it reachable. Adding the Intel prefixes means an Intel keg on an Apple Silicon box without Rosetta is now a live candidate, and that raises `OSError(86, "Bad CPU type in executable")`, which escaped the helper and aborted the analysis. Fixing the exception was part of writing the new candidates correctly.

## Not done

- **P3** `script_analyzers.py:3084` — no diagnostic on the "store found, no matching runtime" path. Still true; `(False, "")` still lands on "Install ILSpy CLI". Worth doing, and now less likely to trigger.
- **P4** returncode 0 with empty stdout; TFM ties resolving in `scandir` order; platform-suffixed TFMs (`net9.0-macos`); the untested Intel prefix; no fallback-level test for `netstandard2.0`.

One P4 is now closed as a side effect: `DOTNET_ROOT` had no assertion, so the line could be deleted with all six tests green. `test_dotnet_root_points_at_the_runtime_parent` covers it, and I verified deleting the line fails it.

## Verification

Five new tests, each mutation-checked — dropping the generic tier, setting roll-forward unconditionally, narrowing `OSError` back, and deleting the `DOTNET_ROOT` line each fail exactly the test that covers them, and only that test. 11/11 in the class. Full `ct_analyzer` suite 30 failed / 101 passed / 14 skipped against a 30 / 96 / 14 baseline: same 30 pre-existing toolchain failures, +5 from this commit. `ruff check` and `format` clean, `make validate` clean.

On your coverage note — the fourth mutation-table row being glob-order dependent was a fair catch. The new tests assert the full ordered call list rather than a single argv, so ordering is pinned rather than incidental.

PR #284, opened 2026-08-27
supply-chain-risk-auditor: don't crash on Windows in the cache owner check

### LINKED ISSUE #273: supply-chain-risk-auditor: collect.py crashes on Windows (os.getuid is POSIX-only)
`supply-chain-risk-auditor` cannot run on Windows. `scripts/collect.py` aborts before collecting anything:

```
File "scripts/sources.py", line 98, in __init__
    if self.cache_dir.stat().st_uid != os.getuid():
                                       ^^^^^^^^^
AttributeError: module 'os' has no attribute 'getuid'
```

## Cause

`sources.py:98`, in `Http.__init__`:

```python
self.cache_dir.mkdir(parents=True, exist_ok=True, mode=0o700)
if self.cache_dir.stat().st_uid != os.getuid():
    raise SystemExit(f"error: cache directory {self.cache_dir} is owned by another user; ...")
```

`os.getuid()` does not exist on Windows, and `os.stat_result.st_uid` is always `0` there, so the check cannot be evaluated as written.

## Why this looks worth fixing rather than documenting as unsupported

- It is the **only** POSIX-only call in the package. `grep -rE 'os\.(getuid|getgid|geteuid|chown|fork)|pwd\.|grp\.|fcntl|resource\.'` over `collect.py`, `sources.py` and `render.py` returns this line and nothing else.
- The package is stdlib-only (`dependencies = []`), so everything else runs unmodified on Windows. Verified by running `python3 collect.py` against six targets in a Linux container: all six collected and rendered cleanly.
- No flag bypasses it — it runs unconditionally in `__init__`. `--cache` changes the path but the check still executes.

## Why a bare `hasattr` guard may not be what you want

The check is a real control: the collector trusts this cache for registry and advisory responses, and a poisoned cache turns a compromised package into a clean "no advisories" verdict. Guarding it with `hasattr(os, "getuid")` silently drops that control on Windows rather than reporting it.

Options that keep the guarantee visible:

1. Skip the check on Windows but emit it as an explicit unassessable/warning line, consistent with how the report already distinguishes "assessed-clean" from "unassessable-with-a-reason".
2. Implement a Windows equivalent — for example, confirm the cache directory resolves under the current user's profile, or compare owner SID via `win32security.GetFileSecurity`.

## Workaround, for anyone hitting this

Run the collector in a Linux container so the check stays meaningful instead of being patched out. Since imports are stdlib-only, `python3` is enough — `uv` is not needed:

```sh
docker run --rm -e GH_TOKEN -v "/path/to/project:/audit:ro" -v "$PWD/out:/out" \
  python:3.13-slim python3 /audit/scripts/collect.py /audit --json /out/findings.json
```

One snag when containerising: `gh_token()` in `sources.py:483` reads the token **only** via `subprocess.run(["gh", "auth", "token"])`, with no environment-variable fallback. A container without the `gh` CLI silently drops to the anonymous 60 req/hour limit, which pushes most repository criteria to unassessable. A `GH_TOKEN`/`GITHUB_TOKEN` fallback would remove the need for a `gh` shim — happy to open that as a separate issue if you prefer.

## Environment

- Windows 10 19045, Python 3.14.6
- Skill installed 2026-08-22 via `npx skills add trailofbits/skills --skill supply-chain-risk-auditor`

kz-tob:
Fixes #273.

## What

`Http.__init__` called `os.getuid()` unconditionally, which does not exist on Windows, so `collect.py` raised `AttributeError` before collecting anything:

```
File "scripts/sources.py", line 98, in __init__
    if self.cache_dir.stat().st_uid != os.getuid():
AttributeError: module 'os' has no attribute 'getuid'
```

Two commits, because review found the reported crash was not the only Windows blocker.

## 1 · The cache-owner check (87d7a66)

The check is a real control rather than a formality. The collector trusts this cache for registry and advisory responses, and a cache another user can write turns a compromised package into a clean "no advisories" verdict.

So this takes **option 1** from the issue: on a platform with no `os.getuid` the check is skipped and `_check_cache_owner()` returns a caveat naming what went unverified, which `collect()` appends to the report notes. It surfaces under **"Method and caveats"** — the channel the report already uses to separate assessed-clean from unassessable-with-a-reason. The POSIX path is unchanged: a foreign owner still aborts.

A bare `hasattr(os, "getuid")` guard was deliberately not used, for the reason the issue gives: it drops the control silently, and the report would then imply a cache it never checked.

**Option 2 (SID comparison via `win32security`) was not implemented.** It needs `pywin32`, and `scripts/pyproject.toml` declares `dependencies = []`. Taking on a third-party dependency to harden a fallback is a poor trade, and the caveat keeps the gap visible either way.

## 2 · Text encoding (fb9346d)

The first commit's description claimed the `getuid` call was the last blocker. That was wrong: it rested on a grep for POSIX-only *calls*, which is a narrower claim than "the run works". No `read_text`/`write_text` in the package passed `encoding=`, so on Windows they decode with the ANSI code page — cp1252, not UTF-8. A `package.json` with a non-ASCII author either raised `UnicodeDecodeError`, which is a `ValueError` but not a `JSONDecodeError` and so escaped `_read_json`'s guards as a traceback, or — where the bytes happened to be cp1252-decodable — was silently mojibaked into the report.

`encoding="utf-8"` now on all twelve text I/O sites, plus the two `subprocess(text=True)` calls, which decode with the locale encoding for the same reason. `_read_json` and `_read_toml` also catch `UnicodeDecodeError` explicitly: JSON is UTF-8 by RFC 8259 and TOML by its own spec, so a file that will not decode is malformed input and earns the same refusal as bad syntax.

## Verification

**Commit 1** — three tests, each mutation-checked:

| Mutation | Result |
|---|---|
| Restore the original unguarded `os.getuid()` | Reproduces the issue's exact `AttributeError` |
| Replace the caveat with a bare `return None` — the shortcut the issue warns against | Fails the test requiring a stated reason |
| Drop the `SystemExit` on a foreign owner | Fails the foreign-owner test |

**Commit 2** — a paired test:

- **behavioural**, running the real collector paths under `-X warn_default_encoding`, where CPython raises `EncodingWarning` on any omitted `encoding=` (PEP 597). This is a property rather than a list of known sites, which is how it caught the two `subprocess(text=True)` calls that no grep had found. The cheaper lever was checked first and rejected: monkeypatching `locale.getpreferredencoding` does not affect `Path.read_text()` on 3.14, so that test would have passed either way.
- **static**, covering what the driver cannot execute, `render.py`'s two sites among them. Confirmed to bite by reverting `render.py:801`.

Also confirmed the caveat survives the renderer: single line, no pipe, passes `_safe_text` and `check_no_forged_lines` — worth checking rather than assuming, since that module exists because interpolated third-party text has forged report structure before.

Full plugin suite 100 passed (was 93). `ruff check`/`format --check` clean. `make validate` clean, version-increment check passes. Version 2.0.0 → 2.0.1.

**Not tested on real Windows.** `monkeypatch.delattr(os, "getuid")` simulates the platform faithfully for that call, and `-X warn_default_encoding` reproduces the encoding defect deterministically on any platform, but a genuine `collect.py` run on Windows is the confirmation this deserves. @izaiascalixto-cmd — if you can try this branch, that would close the loop.

## Known gaps, not addressed here

From review, left open deliberately:

- the same cache-ownership threat aborts on POSIX but produces one bullet at the bottom of the report on Windows, and never reaches stderr — a real asymmetry in how loudly the tool refuses
- the two lines wiring the caveat into `notes` are untested
- `test_cache_owner_is_verified_on_posix` has no skip guard, so this suite fails on the platform it enables
- the interpolated cache path is not backticked, so Markdown can eat backslashes from the one path the reader is told to fix

## Not in scope

The issue also notes `gh_token()` reads the token only via `gh auth token` with no `GH_TOKEN`/`GITHUB_TOKEN` fallback, so a container without the `gh` CLI silently drops to the anonymous 60 req/hour limit and pushes most repository criteria to unassessable. Separate defect; the reporter offered to file it separately, which is worth taking up since it is what makes their own documented container workaround awkward.

Thanks to @izaiascalixto-cmd for an unusually well-scoped report: cause, the "why not `hasattr`" argument, the options, and a working container workaround were all in the issue.


kz-tob:
P2 fixed in fb9346d. The finding was correct and the PR description was wrong — I asserted the `getuid` call was the last blocker on the strength of a grep for POSIX-only *calls*, which is a different claim from "the run works".

## P2 · no `read_text()` passes `encoding=`

`encoding="utf-8"` on all **twelve** sites. You named six; a grep found five more, and a thirteenth only surfaced once the new test ran:

| Site | Named in review |
|---|---|
| `collect.py` `_read_json`, `_read_toml`, `requirements*.txt`, `go.mod` ×2, `render.py` report write | yes |
| `collect.py` `.git/HEAD` + its ref | no — already guarded for `ValueError`, still lacked `encoding=` |
| `collect.py` JSON artifact write | no |
| `sources.py` cache read **and** write | no — both ends of the HTTP cache round-trip, which is where registry text with non-ASCII actually lives |
| `render.py` artifact read | no |
| `sources.py` `gh_token` + pip-audit `subprocess(text=True)` | no — **found by the test, not by me** |

That last row is the interesting one. `subprocess.run(..., text=True)` also decodes with the locale encoding, so it is the same defect wearing different clothes. No grep of mine caught it; `-X warn_default_encoding` did on the first run.

`_read_json` and `_read_toml` now catch `UnicodeDecodeError` explicitly, per your point that it is a `ValueError` but not a `JSONDecodeError` and escaped both clauses. JSON is UTF-8 by RFC 8259 and TOML by its own spec, so a file that will not decode is malformed input and earns the same refusal as bad syntax.

## Tests, deliberately paired

- **Behavioural**: runs the real `collect.py`/`sources.py` paths in a subprocess under `-X warn_default_encoding`, where CPython raises `EncodingWarning` on any omitted `encoding=` (PEP 597). This is a property, not a list of known sites, which is why it caught the subprocess calls. I checked the cheaper lever first — monkeypatching `locale.getpreferredencoding` does **not** affect `Path.read_text()` on 3.14, so it would have been a test that passed either way.
- **Static**: covers what the driver cannot execute — `render.py`'s two sites among them. Verified it bites by reverting `render.py:801`.

Reverting the manifest reader, the go.mod reader, or the `sources.py` cache writer fails the behavioural test; reverting the `UnicodeDecodeError` guard fails the refusal test. Full suite 100 passed (was 96). `ruff check`/`format` clean, `make validate` clean.

## PR description corrected

The "nothing else stands in the way of a Windows run" claim is the part you flagged as unsupportable, and you were right that it would have shipped as the merge commit message. Updated to state what was actually verified.

## Your other findings

- **P3** `collect.py:1394` — the two lines that surface the caveat are untested. Fair; the caveat's *content* is tested but its wiring into `notes` is not. Worth a test that asserts the note reaches the artifact.
- **P3** `sources.py:127` — the same threat aborts on POSIX but yields one bullet at the bottom of the report on Windows, and never reaches stderr. This is the most substantive one left: it is a real asymmetry in how loudly the tool refuses, not a style point.
- **P4** `test_cache_owner_is_verified_on_posix` has no skip guard, so the suite that enables Windows fails on Windows. Embarrassing and correct.
- **P4** the interpolated cache path is not backticked, so Markdown can eat backslashes out of the one path the reader is told to fix — specifically bad on Windows, where the path is the actionable part.
- **P4** `monkeypatch.delattr(sources.os, "getuid")` mutates the global `os` module.
- **P4** three fragments carry an `f` prefix with no placeholder.

None are addressed here; say the word on which you want. The two P4s about the caveat itself (backticking, and the pointless `f` prefixes) plus the missing skip guard are a single small commit if you want them together.

e-q (dismissed):
This looks good, though if we care enough about windows compatibility we should add a windows test runner

kz-tob:
Both P2s fixed in 70b17ce. The first one was mine and it was the bad kind.

## P2 · `test_collect.py:718` — the subprocess half could never fail

Correct, and worse than embarrassing: I wrote a vacuous checker into the very test meant to prevent a vacuous check. `text=True` hits went into `offenders` with the marker, the assertion filtered exactly those into `hard`, and nothing ever read the soft list. The comment even said "check the next line" — it never looked.

I reproduced it before fixing: deleting `encoding="utf-8"` from `sources.py`'s pip-audit call left **four green tests** with the regression live. That is `AGENTS.md`'s "a checker that inspects zero items must fail, not pass", and I shipped it.

Rewritten with `ast`, so a call is seen whole however it is wrapped. Three things fell out of doing it properly:

- **The first AST version was wrong too.** Matching any `.open()` flagged `sources.py:210` — `self.opener.open(request)`, a urllib call returning bytes, where an encoding would be nonsense. Now only the builtin `open` (an `ast.Name`) is flagged, and binary modes are skipped. `Path.open()` in text mode would slip past, which I noted in the docstring: it is one of the reasons the behavioural test exists beside it.
- **Modules are globbed, not listed** — your P3. Same amount of code, strictly better, and the discovery asserts it found at least four.
- **`test_the_encoding_check_detects_a_missing_encoding`** pins both shapes the checker must catch, including the multi-line subprocess form that defeated the line-based version. A guard whose predecessor passed while a regression was live gets its own guard.

Verified: the exact deletion that used to pass now fails with `sources.py:641: run(text=True) with no encoding=`.

## P2 · `collect.py:344` — plain-text readers raise a raw traceback

Correct. `requirements*.txt` and both `go.mod` reads now route through `_read_text`, which refuses with a message naming the likely cause. `_git_commit`'s `except (OSError, ValueError)` was the model, as you said.

Two things worth reporting:

**My first fixture did not reproduce the bug.** `"flask==2.0.0\n".encode("utf-16-le")` is *valid* UTF-8 — nulls are legal — so it decoded cleanly and the test failed with `DID NOT RAISE`. `encode("utf-16")` carries the `0xFF 0xFE` BOM you actually named, which is what fails. Your report was more precise than my reading of it. Incidentally the no-BOM case is the silent-corruption variant: a string full of nulls, no error.

**The third site needed its own test.** `parse_go` short-circuits at the first `go.mod` read, so reverting `_go_indirect`'s read alone left all 103 tests green — and the static check cannot see it, because what is missing there is the guard, not the encoding. Each of the three sites is now individually mutation-verified.

## Verification

104 passed (was 100). Each of the three `_read_text` sites fails when reverted alone; the encoding checker fails on the deletion that previously passed. `ruff` clean, `make validate` clean — which caught something else: my error message quoted a `pip freeze` command, and the modern-python rule forbids documenting pip invocations. Rephrased to keep the actionable part (`Out-File -Encoding utf8`) without it.

## Still open

- **P3** `sources.py:153` — a cache entry with invalid UTF-8 crashes every later run. Probably the most user-hostile one left: it is persistent, and self-inflicted only after a bad write.
- **P3** `sources.py:528` — both subprocess calls now decode strict, so a non-UTF-8 byte from the child escapes their guards. This one is a direct consequence of my last commit, so it is mine to own.
- **P3** all three cache-owner tests fail on Windows, so the suite cannot confirm the behaviour on the platform the PR enables.
- **P3** a UTF-8 BOM is refused as malformed by `_read_json`, and Windows tooling writes one. `utf-8-sig` on reads would take it while still rejecting UTF-16.
- **P4** the check enforcing "name your encoding" omits its own; the caveat names an internal API and fires where the default cache path is already per-user; the pointless `f` prefixes; `_read_json` out of order in the import block.

The two `sources.py` P3s plus the BOM one are arguably a single coherent follow-up about "text that arrives from outside is not necessarily UTF-8", which is the same lesson as this PR one layer out. Happy to take them together.

PR #290, opened 2026-08-29
static-analysis: preserve TypeScript ruleset selection

iflov:
## Summary

- keep JavaScript and TypeScript as separate detection categories
- preserve both `p/javascript` and `p/typescript` for mixed repositories
- add regression coverage proving both scan units and TypeScript include globs survive
- bump static-analysis to 1.4.3

## Problem

The dynamic workflow instructed Detect to collapse a TypeScript-only tree into `javascript` because the JavaScript execution scope also includes `.ts` and `.tsx`. That loses information before Select reads the ruleset catalog, which assigns TypeScript its own `p/typescript` ruleset. The run can therefore complete cleanly without TypeScript-specific rules.

The execution overlap remains intentional; only the detection categories stay distinct.

## Validation

- `node --check workflows/semgrep-scan.js`
- `bash tests/run_scan_tests.sh` — 82 passed
- plugin metadata validation with `--base-ref d1f1575` — no errors


kz-tob (changes requested):
I had Claude dig into the premise here and it couldn't reproduce the problem, at least on the OSS registry.

  Fetching both packs from the registry:

  ```
  p/javascript  74 rules
  p/typescript  74 rules
  ```

  The rule ID sets are identical — `comm` in both directions comes back empty. The 7,400-line raw diff between the two YAML files is serialization order only. So on OSS there are no TypeScript-specific rules in `p/typescript`; it's the same 74 rules under a second name.

  It also checked the language dispatch that the `includes_for javascript` design rests on, with a local two-rule probe:

  ```
  js-only-rule  <-  src/a.ts     # languages:[javascript] fires on .ts
  js-only-rule  <-  src/b.js
  ts-only-rule  <-  src/a.ts     # languages:[typescript] does not fire on .js
  ```

  Together that means a TS-only tree selects the same 74 rules before and after this change, just under a different category name, and a mixed tree now runs those 74 rules over `.ts`/`.tsx` twice. `merge_sarif.py` dedups the copies on `(ruleId, uri, startLine)`, so the cost is scan time rather than doubled findings — but that's the cost #281's text was pointing at, and this removes the warning without showing it was wrong.

  Two things that would change my read:

  1. `rulesets.md:62-63` cites Pro counts of 250+ and 230+, which are different numbers. If the packs genuinely diverge under Pro, that's the case for this change — but then the instruction should be conditioned on Pro, since Detect already establishes OSS-vs-Pro at step 5 and most runs are OSS. Can you share what you're seeing?

  2. `scan-workflow.md` now says a TypeScript-only tree reports TypeScript and not JavaScript. Given the asymmetry above (JS rules run on TS, TS rules don't run on JS), the safe instruction for a TS-only tree is to report both. As written it swaps which pack gets dropped rather than dropping neither — harmless while the packs are identical, but it's the same shape of gap the PR is fixing.

iflov:
Thanks for digging into this. I reproduced the OSS registry comparison: `p/javascript` and `p/typescript` each currently contain 74 unique rule IDs, with no IDs unique to either pack.

I do not have authenticated Pro evidence showing that these two registry packs diverge. The `250+` / `230+` counts in `rulesets.md` were the premise for this change, but they are not sufficient evidence for changing the OSS dispatch behavior. I also agree that the TS-only instruction in this PR is wrong given the language-dispatch asymmetry: if separate packs were required, a TS-only tree would need both JavaScript and TypeScript coverage, not TypeScript alone.

I am closing this rather than adding an unverified Pro-only branch. If an authenticated Pro run later demonstrates different rule-ID sets, this can be revisited with Pro-conditioned behavior and a credentialed regression test. Thanks for the concrete reproduction.

